PR #18, opened 2015-08-12
Add stdin-supplied password support

### LINKED ISSUE #17: Support for stdin-supplied password
Supplying a password via CLI poses a security risk as it might either end up being logged or accessed via the process list, whereas reading from `stdin` (e.g. `echo "myPassword" | veracrypt /path/to/image /mnt/device -p -`) would effectively prevent the latter.

LouisTakePILLz:
Closes #17

I've only tested the changes under Linux, so you might want to keep an eye out for that.

### DIFF
--- a/src/Main/CommandLineInterface.cpp
+++ b/src/Main/CommandLineInterface.cpp
@@ -71,6 +71,7 @@ namespace VeraCrypt
 		parser.AddOption (L"",	L"new-password",		_("New password"));
 		parser.AddOption (L"",	L"new-pim",				_("New PIM"));
 		parser.AddSwitch (L"",	L"non-interactive",		_("Do not interact with user"));
+		parser.AddSwitch (L"",  L"stdin",				_("Read password from standard input"));
 		parser.AddOption (L"p", L"password",			_("Password"));
 		parser.AddOption (L"",  L"pim",					_("PIM"));
 		parser.AddOption (L"",	L"protect-hidden",		_("Protect hidden volume"));
@@ -402,9 +403,21 @@ namespace VeraCrypt
 			Preferences.NonInteractive = true;
 		}
 
+		if (parser.Found (L"stdin"))
+		{
+			if (!Preferences.NonInteractive)
+				throw_err (L"--stdin is supported only in non-interactive mode");
+
+			Preferences.UseStandardInput = true;
+		}
+
 		if (parser.Found (L"password", &str))
+		{
+			if (Preferences.UseStandardInput)
+				throw_err (L"--password cannot be used with --stdin");
 			ArgPassword.reset (new VolumePassword (wstring (str)));
-		
+		}
+
 		if (parser.Found (L"pim", &str))
 		{
 			try
--- a/src/Main/UserInterface.cpp
+++ b/src/Main/UserInterface.cpp
@@ -889,11 +889,19 @@ namespace VeraCrypt
 	{
 		CommandLineInterface &cmdLine = *CmdLine;
 
-		switch (cmdLine.ArgCommand)
-		{
-		case CommandId::None:
+		if (cmdLine.ArgCommand == CommandId::None)
 			return false;
 
+		if (Preferences.UseStandardInput)
+		{
+			wstring pwdInput;
+			wcin >> pwdInput;
+
+			cmdLine.ArgPassword = make_shared<VolumePassword> (pwdInput);
+		}
+
+		switch (cmdLine.ArgCommand)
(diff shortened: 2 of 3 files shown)


idrassi:
My first remark is that in VeraCrypt there is a separation between interactive and non interactive modes. That's why there is a --non-interactive switch and also that's why there is no user interaction during arguments parsing.

In this context, your modification breaks VeraCrypt model since it requests user interactions during arguments parsing and also it doesn't seem to respect the interactive mode switch.

Another point is that there is already an interactive prompt for password when text mode is activated. So, if a user want to provide his password securely, he can simply use text mode.

The only case I can think of is when the user want to launch the GUI and he wants to provide the password interactively instead of typing it in the password dialog or passing it as an argument. 
In my opinion, this case is marginal: the user can simply using VeraCrypt in text mode to mount his volume interactively and after that he launches a GUI instance of VeraCrypt.

Thus, my conclusion is that this modification is not needed.
Any thoughts on my comments?

LouisTakePILLz:
My bad, I had forgotten about this side effect; using the `--stdin` option without any piped data will indeed result in an interactive input, thus causing  the program to halt during parsing.

I had originally intended to allow piping the password via `stdin` so that no user interaction is required.

The following script should give you a pretty good idea of what I'm trying to achieve:

```sh
#!/bin/sh
gvfs-mount -a -d /dev/sda1 2> /dev/null
test "$?" -eq 1 && echo "Can't mount" && exit 1
veracrypt -t -l /media/Storage/path/to/image 1> /dev/null 2>&1 
if [ "$?" -eq 1 ]; then
  zenity \
  --password \
  --title="Unlock device" | \
  veracrypt -t /media/Storage/path/to/image /mnt/device --mount --non-interactive --stdin
  
  # Old (insecure) method
  #veracrypt -t /media/peelz/Storage/.veracrypt/tango /mnt/tango --mount --non-interactive -p=$(zenity \
  #--password \
  #--title="Unlock device")

  test "$?" -eq 0 \
    && zenity --info --title "Mount Device" --text "Successfully mounted the encrypted device." \
    || zenity --error --title "Mount Device" --text "Failed mounting the encrypted device." 
else
  zenity --warning --title "Mount Device" --text "The device is already mounted."
fi
```

LouisTakePILLz:
I made a couple additional modifications so that the `--stdin` option can only be used while in non-interactive mode. The reading operation should now also take place after parsing.

LouisTakePILLz:
@idrassi Any thoughts on the latest changes? I'd make it non-interactive if I could, but that's just how `stdin` works when you don't pipe any data.

idrassi:
Sorry for not reviewing your patch earlier.
The new modification is much better. 
Can you please create a new pull request with only one patch? I don't want to commit this using two patches. A single clean patch is more clearer since the scope of modification is limited. I will commit it right away.

In the future, it would be good a add support for piping keyfiles also instead of passing them in the command line. For this, the current mechanism that you implemented could be extended to define a structured format for data passed through stdin so that we can pass the password alongside a list of keyfiles.

Thank you for your contributions.

LouisTakePILLz:
There you go, that should do the trick.

As for passing keyfiles through `stdin`, I'm not so sure as to whether it is a good idea or not, but surely worth looking into.

idrassi:
Thanks.
As for keyfiles, the idea is to pass the keyfiles names not the content because currently, the names of the keyfiles passed directly in the command line can be spied on much like the password.
For example, we can have a format like (it uses BASE64): pass:valuebase64*keyfile:valuebase64*keyfile:valuebase64

An example:
> pass:dGVzdA==*keyfile:L2hvbWUvam9lL2tleWZpbGUuZGF0

which would mean that the password is "test" and that the keyfile to use is "/home/joe/keyfile.dat"

This would be more flexible.

LouisTakePILLz:
That sounds like a good way to go at it.

Have you considered using a line terminator (or maybe even `\0`) to split the different segments instead of just passing it at the end of the previous Base64-encoded string?

idrassi:
In the specification above, I'm using the * character as a separator. Is there any problem with this?
What is the advantage of using null character instead of *?

I have specifically chosen BASE64 encoding in order to avoid collision with the characters : and * which are used as separators?

LouisTakePILLz:
There shouldn't be any problem with, but in-lining several parameters could be detrimental to the legibility.

Actually, I don't see any reason not to support the three of them. Several GNU Coreutils tools both support splitting on EOL and null characters (e.g. `xargs -0`).

PR #27, opened 2015-10-11
Update Language.zh-cn.xml

Zhangjintao:
Update Simplified Chinese translation.

### DIFF
--- a/Translations/Language.zh-cn.xml
+++ b/Translations/Language.zh-cn.xml
@@ -2,7 +2,7 @@
 <VeraCrypt>
   <localization prog-version="1.16">
     <!-- Languages -->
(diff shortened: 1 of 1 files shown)


idrassi:
Thank you but there is a small mistake: At line 1392, the field "en-name" must not be changed to "zh-cn-name" because it reflects the XML attribute "en-name" present in the header of the file and which identifies the english name of the current language.

Also, there are two fields whose "lang" attribute was changed to "zh-cn" but which were not translated:
- IDC_SHOW_PIM at line 67
- AUTODETECTION at line 827

if these fields can't be translated, then there "lang" attribute should remain "en".

Once these modifications done, I'll commit your pull request.

Zhangjintao:
Thank your correction, I will fix it later.

Zhangjintao:
I have fixed the mistakes.

idrassi:
Thanks. Pull merged.

PR #29, opened 2015-10-19
Update Language.zh-cn.xml

netdcy:
Change some punctuation from English to Chinese.
Minor fixes.


idrassi:
Thanks. Merged although line 1 is marked as changed and line 972 starts with tabs instead of spaces.

Issue #30, opened 2015-11-01
an inquiry about translation


initbar:
When editing the translation of `/Translation/*.xml` files, do you prefer us to directly commit our translations into the pull requests, or through external TrueCrypt resource(s) (i.e. `Crowdin`)?

idrassi:
It is OK to use pull requests but since I can not check the validity of the translation, I ask users to send encrypted email to veracrypt@idrix.fr with information that will help validate that the person in question have indeed the ability to perform the translation.

This is important to avoid having issue and to maintain the quality of the translation inherited from TrueCrypt.

Is this OK for you?

As you can see in the XML files inherited from TrueCrypt, the authors are identified by their real names. In VeraCrypt, anonymous translations can be accepted but the authors should identify themselves through encrypted email and their names will not be put in the XML file.

I started experimenting with Transifex (https://www.transifex.com/idrix/veracrypt/) but I will try Crowdin as well.

initbar:
That sounds good with me :) I've updated it with my real name.

PR #31, opened 2015-11-01
Update Language.ko.xml

initbar:
Linking #30


idrassi:
Can you please explain the reason for modifying existing translations that are inherited from TrueCrypt (e.i. IDM_VERIFY_RESCUE_DISK_ISO & TOKEN_DATA_OBJECT_LABEL)?

initbar:
Yup! - I'll provide explanations for my translations in the following:

#### 14eaa90
<table>
  <tr>
    <td width="33%"></td>
    <td width="33%"></td>
    <td width="33%"></td>
  </tr>
  <tr>
    <td>In this commit, I've took a quick gander and focused on three patterns. The first pattern (above) was areas in which (1) short and medium-length terms|remarks could be translated without much hardship. One of the most important part was keeping the terms consistent with those used by mister Kim Young without fragmenting all over the place with similar synonyms.</td>
    <td>(2) Keeping the general structure by removing the space after colon.</td>
    <td>(3) trailing whitespaces.</td>
  </tr>
</table>
#### 3eb383d
<table>
  <tr>
    <td width="33%"></td>
    <td width="33%"></td>
    <td width="33%"></td>
  </tr>
  <tr>
    <td>I've reverted my mistake <code>~</code> back to <code>&lt;-&gt;</code> after checking that other language had used <code>&lt;-&gt;</code> symbol.</td>
    <td>Translation of longer* sentences.</td>
    <td>Correcting <code>lang=""</code> attribute values.</td>
  </tr>
</table>
Now up to this point, mister Kim's translations weren't touched for the following final revision (commit):
#### 6691cc0
Initially, `IDM_VERIFY_RESCUE_DISK_ISO` and consequent `Favorites` translations were in English (except the last element):



In commit `14eaa90` line 264, I've translated to `응급복구 디스크 ISO 완전성 검증` - which later I've realized that using `완전성`, which means "holistic, on the whole, integrity" would be redundant and aesthetically unappealing due to inconsistency (even with and without the extra 3 characters, users would still be able to correctly infer):



As for `TOKEN_DATA_OBJECT_LABEL`, mister Kim's translation is "technically" correct, as the literal translation is: "file name". However, in terms of tokens, I've revised to `토큰 레이블`, which reads "token label". I thought this would be less confusing.

In `LABEL_ITEM`, mister Kim's translation was `라벨` - which does translate to "label", but reads "lah-bell". I preferred if it was consistent with the terms used in other places, and changed to `레이블` (which literally reads "label"). In a minor note, in `IDT_FAVORITE_LABEL`, he uses `레이블` instead of `라벨`.

The `IDT_NUMBER_KEYFILES` to `IDT_KEYFILES_BASE_NAME`'s removal of `의` is because (1) the `의` pushes the following characters out of the windows and hide the them from view. (2) `의` just (roughly) means: "of". The overall translation is much more clean to remove this and directly say:

| from | to |
|:------:|:---:|
| `키파일 수:` | ~Keyfile numbers: |
| `키파일 사이즈 (Byte):` | ~Keyfile size (Byte): |
| `키파일 배이스 이름:` | ~Keyfile basename: |

So after this revision, it looks like:

PR #44, opened 2016-01-22
Language files: Update German translation

makomi:


makomi:
main points
* make the text easier to comprehend
* use quote types consistently
* grammar, typos, ...

idrassi:
Thank you for your contribution. You definitely worked hard to review all translations!

One problem introduced by your changes is that some buttons have their text truncated because you added a space between the text and "...".. For example, the password dialog has the button "Schlüsseldateien... truncated:
VeraCryptGermanTruncated

So, I would like that you remove all introduced spaces in XML "control" nodes because of the size constraint. All, it is important to ensure that for these XML nodes the text size doesn't grow because the current translations take into account this size constraint.

For the other changes, they depart from the current translation style so I have to consult with the current German translator to have his comments. Since I don't speak German, I rely on trusted external people for such review. I will update you once I have his input.

makomi:
Thank you for the detailed feedback! In the updated commit, I removed all introduced spaces in XML "control" nodes and kept those in the "string" nodes.

Is there a style guide available for editing the translation files? If not, I could help create that guide. Such a guide would help reduce the unwanted edits.

By the way: I only reviewed the translations up to "HIDVOL_PROT_WARN_AFTER_MOUNT_PLURAL". All changes after that entry are only for consistency's sake.

makomi:
I amended some things, so please look at d75a489.

idrassi:
The usual translator confirmed that indeed your changes bring several enhancements, so I would like to merge your pull request.

But I have made some changes to the order of some XML nodes and this will cause an issue if I merge your pull request. Since your last changes:

- Node "INIT_DLL" was replaced by "INIT_RICHEDIT" and "INIT_DLL" is displayed to the bottom of the file
- The following node were displaced to the bottom: "IDC_BOOT_LOADER_CACHE_PIM", "IDC_PREF_CACHE_PIM", "IDC_SHOW_DISCONNECTED_NETWORK_DRIVES", "PASSWORD_UTF8_TOO_LONG", "PASSWORD_UTF8_INVALID".
- New node "ERR_EXFAT_INVALID_VOLUME_SIZE" added at the bottom

Can you please update your pull request to adapt to these changes? 

Thank you in advance.

idrassi:
Just in case, I have created an XML file that combines your changes and the modifications I did recently. Here it is: [Language.de.zip](https://github.com/veracrypt/VeraCrypt/files/109079/Language.de.zip)

It can help for your pull request (if I didn't do a mistake!)

makomi:
I am looking forward to updating my pull request. However, I won't be at my computer until Monday evening.

makomi:
I amended the commit and fixed two translation related things in two of the new nodes.

BTW: Your XML file was flawless. I used your instructions to modify my original commit and compared the result to your version.

idrassi:
Thank you for the update.

Actually, I can't merge your pull request because of conflicts...I don't know exactly why but it looks like it is caused by the fact that I added a new  field yesterday at the bottom of the XML file called IDT_ENTROPY_BAR.
Can you please propose a translation for it and create a new pull request?

Hopefully this time I'll be able to merge it.

makomi:
Sure thing, all done!

idrassi:
Thank you for your contribution!

Issue #53, opened 2016-03-14
Adhere to the XDG Base Directory Specification for configuration file paths


davidfoerster:
## The current state
Currently VeraCrypt stores its configuration files in a hard-coded, platform-dependent path. While this path follows reasonable defaults on Windows and OS X, it contributes to the clutter of the user home directory on Linux. The preferred default configuration file location of modern Linux systems is underneath `~/.config/`.

## Change request
I suggest you change the default configuration file location to depend on the environment variable `XDG_CONFIG_HOME` as suggested by the [XDG Base Directory Specification]. In its absence you can fall back to the current defaults on Windows and OS X and to `~/.config/` on Linux and other *nices.

## Rationale
This would allow users and system operators to specify different user configuration file paths to their convenience and according to their system configuration ([example use case](//askubuntu.com/a/743466/175814)).

## Considerations
As for backwards compatibility on Linux, in the absence of `$XDG_CONFIG_DIR/VeraCrypt`, look for an existing `~/.VeraCrypt` directory and upon success use latter as the per-session configuration path; otherwise use the former.

(Migrated from [CodePlex](https://veracrypt.codeplex.com/workitem/408) because I overlooked the active GitHub mirror even though I'm here more often.)

[XDG Base Directory Specification]: //specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

davidfoerster:
This seems to be a restriction of [`wxStandardPaths::GetUserDataDir`](http://docs.wxwidgets.org/trunk/classwx_standard_paths.html#a4752213ef4d7bdc71170b9e5c3691f0f) which VeraCrypt uses to determine the user configuration path (see [`src/Main/Application.cpp:53`](../blob/dc1593d60f63aa951c1463ade13c97cfde94d2f5/src/Main/Application.cpp#L53)).

PR #55, opened 2016-03-31
Use XDG_CONFIG_HOME to determine the path of the configuration

davidfoerster:
Adhere to the [XDG Desktop Specification] and use the environment variable `XDG_CONFIG_HOME` to determine location of configuration files on *all* platforms. If it is unset or empty resort to platform-specific defaults.
    
On Windows and OS X, `wxStandardPaths` provides correct defaults (equal to the previous hard-coded paths) but on Linux and other Unices `~/.config/appinfo` would be better than `~/.appinfo`. This means we treat those platforms as a special case. It also means that we may need to fall back to the legacy location if it exists but the new location doesn't.

Unfortunately I have no way to test this on OS X or Windows, though the algorithm and code for these platforms are much more straightforward than on Linux.

See issue #53.

  [XDG Desktop Specification]: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html


idrassi:
I agree with the changes but we need extra code handle existing configuration files: for users who are already using VeraCrypt, deploying this change will result for them losing there favorites list and their various other configurations.

An idea would to keep the original code in a method (for example `Application::GetConfigFileLegacyPath`) and in the new implementation we will look if the requested file already exists in the legacy path and if yes we will move it to the new XDG path before returning from the function.

Can you implement this extra logic in your patch? it is the only missing thing in order to merge this pull request.

Thank you again for your help.

davidfoerster:
> An idea would to keep the original code in a method (for example `Application::GetConfigFileLegacyPath`) and in the new implementation we will look if the requested file already exists in the legacy path and if yes we will move it to the new XDG path before returning from the function.

I already implemented [something very similar](/davidfoerster/VeraCrypt/blob/xdg-config-paths/src/Main/Application.cpp#L56) but it's not very obvious since the changes only affect Linux and BSD users. The paths on Windows and OS X stay the same unless the user runs VeraCrypt with `XDG_CONFIG_HOME` set to something different than `%APPDATA%` (Windows) or `$HOME/Library/Application Support` (OS X).

Let me know if you have suggestions for a different behaviour for the look-up of legacy configuration paths or  would like some refactoring.

idrassi:
Indeed, I missed this part.
My idea was to migrate the file from the legacy path to the standard path. I see that your approach is more prudent in that it simply points to the existing file. This is acceptable for me.

One remark about the static pointer configDirC: what about replacing it with a wxScopedPtr<wxString*>? It will be much cleaner than using a static pointer.

davidfoerster:
> One remark about the static pointer configDirC: what about replacing it with a `wxScopedPtr<wxString*>`? It will be much cleaner than using a static pointer.

I can do that if you want to. I'm not sure if it'll do much good though. At least a few years ago neither C++03 nor the documentation of most C++ compilers were clear when the destructors of static values are called; so the recommendation for static values was to just use a pointer to a heap-allocated object instead of the object itself – at least that way it would be clear that they are never destroyed unless explicitly requested by the programmer.

While looking for a source I found https://stackoverflow.com/a/2204628/2461638, so I'm going to go with your suggestion of a scoped pointer.

idrassi:
Thanks!

PR #63, opened 2016-05-06
update language.hu

falu:
some hungarian translations


idrassi:
Following some changes in the XML files, there are conflicts with your pull request.
Is it possible to solve them?

PR #66, opened 2016-05-13
Makefile: fix build under gcc6

FadeMind:
Veracrypt for build under GCC6 required using compatibility flag `-std=c++98` against change default C++ frontend in GCC6.

The C++ frontend now defaults to C++14 standard instead of C++98 it has
been defaulting to previously, for compiling older C++ code that might
require either explicitly compiling with selected older C++ standards,
or might require some code adjustment, see
http://gcc.gnu.org/gcc-6/porting_to.html for details.  The experimental
C++17 support has been enhanced in this release.

from: https://gcc.gnu.org/ml/gcc-announce/2016/msg00000.html

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -40,7 +40,7 @@ export AS := nasm
 export RANLIB ?= ranlib
 
 export CFLAGS := -Wall
-export CXXFLAGS := -Wall -Wno-unused-parameter
+export CXXFLAGS := -Wall -Wno-unused-parameter -std=c++98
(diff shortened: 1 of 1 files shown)


idrassi:
I have already manually fixed gcc-6 errors without resorting to the c++98 switch but many warnings are displayed: https://github.com/veracrypt/VeraCrypt/commit/942cf1635a88bd099d9f6108d8c42c15fa0dee0c

This c++98 compatibility switch introduces errors when linking dynamically against the wxWidgets provided by the system since this library is built using C++14. 

Thus, i'm reverting your pull request.

PR #69, opened 2016-05-20
Added two line to enable support for retina ui on retina Macs.

anarchyonline:
Currently Veracrypt ui is "low resolution" only when installed on Mac.
Adding these two lines enables Veracrypt to run as a retina app.


idrassi:
Thanks.
I can't test this on my side but I will merge the pull request since it seems be a straightforward modification.

PR #74, opened 2016-06-23
correct size unit in console help MB -> MiB

oceanBT:


csware:
cf. bb123bb3e8719d012884d144d326e5e8c8c15786

PR #118, opened 2016-11-28
Document --truecrypt option

pnijjar:
There are a number of commandline options that are not documented in src/Main/UserInterface.cpp but which exist in src/Main/CommandLineInterface.cpp . I installed the text-only version of VeraCrypt and got caught because I did not know there was a --truecrypt option to make opening existing TrueCrypt shares easier. 

I do not know whether GitHub is okay for submitting requests, and I did not set up the build environment to test whether this patch compiles. But I thought I would submit this patch in case it is helpful in developing a proper one.


pnijjar:
Note that I did some rewording from the description of the option in the manual, and I made up a story about which of --text and --truecrypt should be first (since the source suggests that both want to be the first option specified). You will want to check this before accepting the pull request as-is.

idrassi:
Thank you for your contribution. Indeed, the -tc option was missing. I'll merge your pull request.

PR #122, opened 2016-12-20
Use the user provided PKG_CONFIG_PATH instead of pkg-config in makefiles

gktrk:


idrassi:
The proposed modification is wrong because `PKG_CONFIG_PATH` contains a list of directories that the pkg-config script checks for certain files it can use to do its job (e.g. the directory /usr/lib/pkgconfig) and not the path to `pkg-config` binary. You can read more about it at https://dev.gentoo.org/~mgorny/pkg-config-spec.html#environment-variables

PR #166, opened 2017-06-07
Graphic changes

tkrn:
Graphic updates. Please review that they appear correct upon compiling.


idrassi:
As you may expect, there are some issues like the screenshot below.
Some questions:

- why did lower the dimension of icons from 256x256 to 128x128?
- changing the size of bitmaps create issues in the UI. Is it possible to keep the same size?

Thanks.


about_screenshot

tkrn:
Idrassi - Thanks for catching the resolution issue. It's missing the 512x512 version of it. It has the 256x256 & the 128x128 copies. Secondly, I'm in the process of of working on correcting the display on the form so it fits correctly. Let me work on these two items and let me get back to you. Thanks,

tkrn:
Idrassi -

1 - I updated the icons to be 256x256 and smaller recommended versions for Windows builds.
2 - Modified the Common.rc file to take in account the new size of the About Logo Banner (see below).

Let me know if there are questions.

Issue #173, opened 2017-06-25
An option to mount a Encrypted volume


mynameherebro:
An option to mount a Encrypted volume from the windows left click menu on both the file and the drive letter much the way that bitlocker does for windows would make it much easier for mounting and dismounting volumes.

PR #174, opened 2017-06-27
Context menu feature

tkrn:
In response to Issue #173 An option to mount a Encrypted volume.

Right-click Mounting:

Administrator privileges:

No administrator privileges:


idrassi:
Thank you for sharing your work. This is an interesting feature.
The pull request contains patches not related to the context menu feature. I prefer the pull request to contain only commits related to the implemented feature.

Concerning the implementation itself, it is not enough to be administrator to be able to write to the registry in order to activate the context menu. You will have to take UAC into account and your code doesn't handle this since it supposes that it is enough to be administrator in order to write to HKEY_CLASSES_ROOT registry key. 

In VeraCrypt, we already handle UAC through the use of COM objects that exposes the needed interfaces and we launch elevated instance of the current process in order to be able to execute the COM object's method with full administrative privileges. 
You can take a look at [SetDriverServiceStartType](https://github.com/veracrypt/VeraCrypt/blob/master/src/Mount/MainCom.idl#L42) method in https://github.com/veracrypt/VeraCrypt/blob/master/src/Mount/MainCom.idl and see how it is used in [BootEncryption::RegisterBootDriver](https://github.com/veracrypt/VeraCrypt/blob/master/src/Common/BootEncryption.cpp#L4718).

Is it possible to update your code in order to use this existing UAC aware mechanism?

Thank you.

tkrn:
Thanks for the guidance. I will update the code to take account UAC + elevated admin privileges. re-do a pull request without the graphic changes which got wrapped into the pull by accident.

PR #214, opened 2017-09-29
Linux: workaround to mount system-encrypted partition on nvme drive

gv5470:
The naming scheme of Linux block devices representing partitions hosted on NVME drives differs from the usual (e.g. /dev/nvme0n1p2 instead of /dev/sda2). This prevents mounting of Windows encrypted system partitions from Linux, since the name of the hosting drive, which is required in this case, is not correctly deduced by VeraCrypt. Therefore I suggest this simple workaround, maybe not ideal but at least it doesn't seem to cause any harm.


gv5470:
I now have a better solution, so I'm closing this one.

PR #242, opened 2017-11-27
Fix TrueCrypt mode checkbox is always set

qncm:
BUG: "TrueCrypt mode" checkbox in the Mount Dialog is always set despite the state of the corresponding flag in Default Mount Options.

### DIFF
--- a/src/Main/Forms/MountOptionsDialog.cpp
+++ b/src/Main/Forms/MountOptionsDialog.cpp
@@ -38,7 +38,7 @@ namespace VeraCrypt
 		PasswordPanel->SetCacheCheckBoxValidator (wxGenericValidator (&Options.CachePassword));
 		
 		if (options.Path && options.Path->HasTrueCryptExtension() && !disableMountOptions 
-			&& !options.TrueCryptMode && (options.Pim <= 0))
+			&& options.TrueCryptMode && (options.Pim <= 0))
 		{
 			PasswordPanel->SetTrueCryptMode (true);	
 		}


idrassi:
Your statement is not correct: if the container file has ".tc" extension, then the TrueCrypt Mode checkbox will be set regardless of the default Mount Options. But if the file container don't have ".tc" extension, then the if statement will be false and its code will not be executed, thus the default Mount Options will be used.

I think you missed `options.Path->HasTrueCryptExtension()` in the if statement. This if block was added to explicitly handle TrueCrypt file container so that TrueCrypt Mode is always set for them (see commit https://github.com/qncm/VeraCrypt/commit/1ef6177ae3605b4aa24beb6d9a0da214c15e100e#diff-4a442cf13f91744f9883bf33cb61bab3).

Did I miss something?

idrassi:
Do you have a use case that exhibit the problem you are talking about?
It will be quicker because I don't see any bug in VeraCrypt logic for the handling of TrueCrypt Mode: If default TrueCrypt Mode is checked, it will be always set and if it is not set as default, it will be checked **only for files that have .tc extension**.
The code do exactly that: the constructor of `VolumePasswordPanel `sets TrueCrypt Mode checkbox according to default mount options and the if block is only executed if `options.TrueCryptMode` is false (thus the ` !options.TrueCryptMode` in the if statement which you reversed).

qncm:
The file I'm trying to mount has  .tc extension and VeraCrypt format at the same time. Plus I didn't understand the logic of code correctlу. I should have studied it more closely before proposing the patch. Sorry for inconvenience.

PR #250, opened 2017-12-26
Update Language.de.xml

Atalanttore:
I've translated all new strings for the upcoming version 1.22 and improved one existing string.


idrassi:

PR #267, opened 2018-01-21
Fix wxWidgets assertion failed when backing up/restoring volume heade…

### LINKED ISSUE #100: Assertion failed while backing up and restoring volume header
I'm running VeraCrypt 1.18 on Ubuntu Mate 14.04.5.
I needed to backup the volume header of a container; after typing the outer volume password, I got the following error:

```
An assertion failed!
 ../src/generic/choicdgg.cpp(511): assert "sel >= 0 && (unsigned)sel < m_listbox->GetCount()" failed in SetSelection(): Invalid initial selection

```

After selecting **Continue**, the backup operation ended successfully.

To check the backup integrity, I restored immediately the backup on a copy of the same container; the operation has been completed but I got the exact same error.

In the following file you can see the backtrace error [backtrace.txt](https://github.com/veracrypt/VeraCrypt/files/490514/backtrace.txt)

gv5470:
…r (closes #100)


idrassi:
Originally, that code used -1 as a parameter to wxSingleChoiceDialog::SetSelection as a way to avoid any default selection but this hack stopped working in recent versions of wxWidgets anyway.
Your patch restores default behavior.

PR #288, opened 2018-03-06
Update language.ro.xml to 1.22-BETA4

TigerxWood:
Corrected


idrassi:
Their was a syntax error in the translation of the field IDC_BOOT_LOADER_CACHE_PIM: the ampersand symbol was not properly escaped, it was written as &ampP; instead or &amp;P.
I fixed this and I pushed corrected Romanian XML file.

PR #318, opened 2018-05-03
Added line breaks between copyright holders

sandakersmann:
Added line breaks between copyright holders


sandakersmann:
@idrassi Does this look good?

PR #381, opened 2018-11-20
Do not create uninstall shortcut in startmenu

csware:
Creating such a start menu entry is a leftover of the ancient Win 3.x time where there was no central control panel for removing programs.

Also see the Windows guidelines, where creating an uninstall shortcut is discouraged: https://msdn.microsoft.com/en-us/library/ms954377.aspx


idrassi:
Thank you!

PR #389, opened 2018-12-03
Information on the translation process

sandakersmann:
Information on the translation process


sandakersmann:
@idrassi I think it would be nice to add this info for contributors.

PR #453, opened 2019-05-03
Some cleanup related to "Invalid characters..." on mount issue.

karzhenkov:
This is another solution, perhaps a bit more elegant than existing one (646679da4d79bf7f8af22c44c7ae8498086a88a6). Also, here is a bit more interesting to consider.

Existing solution is a workaround for some mysterious gcc-5  STL implementation issue. Meanwhile, actual cause of the problem is that gcc-4 STL implementation detail was exploited.

Older versions of gcc STL use "copy-on-write" optimization for `std::string`. Such optimization was permitted by C++ standard but not required. Under some conditions an old `std::string` behaves as a shared pointer to internal string representation.

The `foreach` macro makes a copy of a container. This copy is destroyed immediately after the iteration is completed. Pointers to char retrieved from copied strings become invalid just after the loop. In older gcc STL implementations, however, they can be used as long as there are other instances of `std::string` that share the internal storage with copied and destroyed strings. From the C++ standard standpoint dereferencing of such pointers is (and always was) an "undefined behavior". Now the standard doesn't permit "copy-on-write" strings; unfounded assumptions are broken.

Test application "ForEachTest" to demonstrate situation described can be found here: 16edc6c053b2fb47cc0605f71fca1ada0d184ffc. With gcc-8 it gives the following:
```
create container
0x55e2411e9ea0 ->   1  create item
0x55e2411e9ee0 ->   2  create item
0x55e2411e9f20 ->   3  create item

get data pointers
0x55e2411e9f60 -> 101  copy item
0x55e2411e9fa0 -> 102  copy item
0x55e2411e9fe0 -> 103  copy item
0x55e2411e9f60 -> 101  get valid pointer
0x55e2411e9fa0 -> 102  get valid pointer
0x55e2411e9fe0 -> 103  get valid pointer
0x55e2411e9f60 -> 101  destroy item
0x55e2411e9fa0 -> 102  destroy item
0x55e2411e9fe0 -> 103  destroy item

use data pointers
0x55e2411e9f60 ->   0  is pointer valid?
0x55e2411e9fa0 ->  64  is pointer valid?
0x55e2411e9fe0 -> 128  is pointer valid?

destroy container
0x55e2411e9ea0 ->   1  destroy item
0x55e2411e9ee0 ->   2  destroy item
0x55e2411e9f20 ->   3  destroy item 
```


idrassi:
Thank you for this better fix and for the explanations.

PR #454, opened 2019-05-03
Fix issues related to "foreach" macro

karzhenkov:
Proposed alternative solution of "Invalid characters..." issue on mount (#453) is not to use the "foreach" macro in one particular case. This macro is used in many other places and
-   is inefficient (makes unnecessary copies)
-   is unsafe in some sense (see #453)
-   has a safe and efficient equivalent in the "new" C++ standard (can we still consider C++11 new?))

Here is yet another variant based on "range for" operator (requires C++11). The "friends" of the "foreach" macro with reversing/dereferencing cannot be expressed by "range for" directly, so they use simple container adapters.

### DIFF
--- a/src/Platform/ForEach.h
+++ b/src/Platform/ForEach.h
@@ -13,110 +13,73 @@
 #ifndef TC_HEADER_Platform_ForEach
 #define TC_HEADER_Platform_ForEach
 
+#include <iterator>
+#include <utility>
+
 namespace VeraCrypt
 {
-	class ForEach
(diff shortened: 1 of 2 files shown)


idrassi:
Thank you again, this is indeed a more cleaner implementation but for now we need to support compilers that don’t come with C++11 so I’m not going to merge this for the moment.

karzhenkov:
Ok, maybe sometime later) I hope that someday С++11 will be generally accepted.

PR #486, opened 2019-09-12
Added creation of Rescue USB Drive

Van-Fouran:
- In my tests it was not possible to create a usb Stick with Unetbootin like mentioned in this post: https://sourceforge.net/p/veracrypt/discussion/features/thread/5453c652/#97df
- I followed the instructions mentioned by Mounir IDRASSI in this post: https://sourceforge.net/p/veracrypt/discussion/features/thread/5453c652/#8e81
- In my test the usb_format.exe does not work, but the Windows internal formatting led to the same results.


idrassi:
Thank you for this contribution.

PR #532, opened 2019-10-30
Start following IEEE 1541 agreed naming of bytes

Meteor0id:
This standard exists to prevent any confusion about the actual number of bytes. It has been agreed on by experts and is being used more widespead these day. Let's start properly naming the number of bytes, which is absolutely important in disk encryption software.

initial commit only addresses language.xml. Contributions are welcome to cover the other strings in other files.


bazzilic:
Are you sure that every KB you replace with KiB meant kibibytes and not kilobytes?

Meteor0id:
I checked the best I could, there wasn't a single case where bytes were expressed as multiple of 1000 instead of 1024. But reviews are welcome. Whoever does just comment to confirm or correct these changes.

idrassi:
I agree that it is time for this naming change but before accepting this pull request, I want to query about 2 lines in language.xml that seem to be modified without reason. Any thoughts on this?

Meteor0id:
Do note that there could still be a couple strings in other sources not converted, and that the html doc files also haven't been converted. A second PR would make sense to complete the job.

PR #539, opened 2019-11-10
Windows: Allow system encrypted devices to be mounted from WindowsPE - even if changing the keyboard layout failed

csware:
When my system broke and I had to use WindowsPE to rescue some files the password dialog always closed immediately. There was no chance to mount the system device using the GUI. It took me a while to realize that I could try using the CLI by passing the password as a parameter (which might not be so obvious for not advanced users).

Please beware: I was not able to test this change (I could not build VeraCrypt).


idrassi:
Thank you for reporting this issue and for the  workaround. It is indeed problematic to block GUI mounting because of keyboard layout switching problem and I agree with your new approach.

PR #559, opened 2019-11-28
fixed typo "This [is] not a bug in VeraCrypt"

vstoykovbg:
Original text: `This not a bug in VeraCrypt but a limitation of the operating system.`
Corrected: `This is not a bug in VeraCrypt but a limitation of the operating system.`

I am not sure if there should be a comma (", but a limitation...").


idrassi:
Thank you!
As for the comma, it is not needed at that location.

PR #561, opened 2019-12-01
UNIX: make sector size mismatch error more verbose (#552)

### LINKED ISSUE #552: ParameterIncorrect at VeraCrypt::CoreUnix::MountVolume:477
Some years ago I created a whole disk encrypted drive with the latest working Truecrypt 7.1a using Seagate Barracuda 3Tb SATA HDD in its own (factory-manufactured) external USB3 enclosure "Seagate expansion desktop drive" in Linux.  The USB controller of the external enclosure is broken now and this SATA drive is not detected by the system in this enclosure.  I have another identical enclosure and after inserting the SATA drive into the new one  I can mount the encrypted drive with Veracrypt 1.23 successfully.   To make a backup of this drive I bought simple Yottamaster USB 3.0 Docking Station https://www.amazon.de/gp/product/B0722Z5C4J/ and tried to mount the same encrypted SATA disk using this docking station.  Unfortunately, I got an error
> ParameterIncorrect at VeraCrypt::CoreUnix::MountVolume:477

from Veracrypt in Linux.
In dmesg log I see  one interesting warning for attached drive: 

> [68597.715727] sd 6:0:0:0: Attached scsi generic sg2 type 0
> [68597.716756] sd 6:0:0:0: [sdc] 5860533168 512-byte logical blocks: (3.00 TB/2.73 TiB)
> [68597.716762] sd 6:0:0:0: [sdc] 4096-byte physical blocks
> [68597.717123] sd 6:0:0:0: [sdc] Write Protect is off
> [68597.717128] sd 6:0:0:0: [sdc] Mode Sense: 5f 00 00 08
> [68597.717662] sd 6:0:0:0: [sdc] Write cache: enabled, read cache: enabled, doesn't support DPO or FUA
> **[68597.718112] sd 6:0:0:0: [sdc] Optimal transfer size 33553920 bytes not a multiple of physical block size (4096 bytes)**
> [68597.739981] sd 6:0:0:0: [sdc] Attached SCSI disk

If I understand that info correctly, Linux kernel assigns now a different block size for this drive? Was original block size encrypted within Truecrypt header? Is this a bug of docking station/ Linux kernel or do we have a problem within Veracrypt code? Should I return this docking station? I mounted another small  500 Gb SATA SSD without problems using this docking station.

alt3r-3go:
Here's the PR to increase the verbosity of the error message we throw upon device and VC volume size mismatch, to hopefully make it more self-explaining and actionable.

Tested manually on Linux.

Fixes #552 

dev-vol-sector-size-mismatch

### DIFF
--- a/src/Core/Unix/CoreUnix.cpp
+++ b/src/Core/Unix/CoreUnix.cpp
@@ -473,8 +473,10 @@ namespace VeraCrypt
 
 		if (options.Path->IsDevice())
 		{
-			if (volume->GetFile()->GetDeviceSectorSize() != volume->GetSectorSize())
-				throw ParameterIncorrect (SRC_POS);
+			const uint32 devSectorSize = volume->GetFile()->GetDeviceSectorSize();
+			const size_t volSectorSize = volume->GetSectorSize();
+			if (devSectorSize != volSectorSize)
+				throw DeviceSectorSizeMismatch (SRC_POS, StringConverter::ToWide(devSectorSize) + L" != " + StringConverter::ToWide(volSectorSize));
 		}
 
 		// Find a free mount point for FUSE service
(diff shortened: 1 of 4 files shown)


idrassi:
Thank you!

PR #568, opened 2019-12-08
Linux: fix NOASM compilation (#563)

alt3r-3go:
Here's the proposed fix - looks like it was the introduction of the optimized Camellia/Twofish cipher implementations that broke it (as there were no compile guards added for the `NOASM` case).

To the best of my understanding (which is admittedly yet incomplete), the proposed fix is correct, but I'll be glad to take any additional advice and correct if necessary. The main idea is to use a proper `define`, which indicates no Assembler support being enabled (and not only 64-bit, as it currently is) in the `Makefile` and then propagate respective `if/if defined` guards where appropriate. There's one specific place of which I'm not 100% sure whether it's indeed just 64-bit-specific, or Assembler-optimized-implementation-specific - [this one](https://github.com/veracrypt/VeraCrypt/compare/master...alt3r-3go:gh563-fix-noasm-compilation?expand=1#diff-601db8dd08f7ba9228a21d8c4e042b7eR25), as I haven't had enough time to analyze it in detail, however initial code review + benchmarks seem to show it's the latter, so it _is_ in the PR.

Tested on Linux by compiling with `NOASM=1` and without (both now compile okay), then running all benchmark suites (encryption, PKCS-5 PRF, hash) and observing no errors and expected performance drop with `NOASM=1`.

Fixes #563.

PR #569, opened 2019-12-09
Add burn calls for temporary ss variable

hannob:
The functions aes_{en,de}crypt_key{128,192,256} in Aeskey.c use a temporary variable ss that contains key material. To make sure the key bits don't stay in memory we should add burn calls.

This code only gets used with NOASM builds on modern systems, using patch from https://github.com/veracrypt/VeraCrypt/pull/568 to make it build.


idrassi:
Thank you!

PR #570, opened 2019-12-20
Fix the compilation against WxWidgets when NOGUI=1

gktrk:
VC fails to build with system wxGTK compiled with '--disable-gui':

```
In file included from TextUserInterface.cpp:27:
TextUserInterface.h: In member function ‘virtual bool VeraCrypt::TextUserInterface::Initialize(int&, wxChar**)’:
TextUserInterface.h:50:63: error: ‘wxAppBase’ has not been declared
   50 |   virtual bool Initialize (int &argc, wxChar **argv) { return wxAppBase::Initialize(argc, argv); }
```

wxAppBase is only defined when wxGTK is built with `--enable-gui`, which is unlikely with headless systems. Therefore, revert the commits that introduced the breakage, simplify the WX_CONFIG_LIBS logic, and pass `-DwxUSE_GUI=0` when `NOGUI=1`

### DIFF
--- a/src/Main/Main.make
+++ b/src/Main/Main.make
@@ -81,7 +81,15 @@ CXXFLAGS += -I$(BASE_DIR)/Main
 
 #------ wxWidgets configuration ------
 
+ifdef TC_NO_GUI
+ifdef VC_WX_STATIC
+WX_CONFIG_LIBS := base
+else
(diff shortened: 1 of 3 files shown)


idrassi:
Thank you for this contribution.

gktrk:
Thank you for merging it quickly.

PR #581, opened 2020-01-27
 Adding Processor Groups support for more than 64 processors

techvintage:
Added support for Processor Groups to enable more than 64 logical processors when spawning threads for encryption/decryption algorithms.This should give considerable speed up for encryption/decryption time.

### DIFF
--- a/src/Common/EncryptionThreadPool.c
+++ b/src/Common/EncryptionThreadPool.c
@@ -10,13 +10,19 @@
  code distribution packages.
 */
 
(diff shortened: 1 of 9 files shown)


idrassi:
This PR also changes the project files to use Visual Studio 2019. Is possible to amend the PR to only include the code change without project files or solution modifications?

techvintage:
As requested,Updated the pull request to include only source code change stripping VS2019 files.

PR #619, opened 2020-04-26
Update README.md

alishir:


alishir:
At that time I added for more readability, but now I checked the readme and the commands are in a code block.

PR #649, opened 2020-06-28
Hide WaitDialog when "/silent" in use

### LINKED ISSUE #648: Command line "Silent" argument does not work in "VeraCrypt Format.exe"
Hi.

according to the "[Command Line Usage](https://www.veracrypt.fr/code/VeraCrypt/plain/doc/html/Command%20Line%20Usage.html?id=4a7ed6926b276cf88c2485ea2a5e95b7f91e4ac1)" I've added `/silent` to the `VeraCrypt Format.exe /create` command, but the following screen appeared:
VeraCrypt-Loading

I've tested with both `VeraCrypt Format.exe` and `VeraCrypt Format-x64.exe` and both produced the same issue.

Just an FYI - This is different behavior than the other exe (`VeraCrypt.exe`) where `/silent` does work as expected.

Thanks.

itsho:
Fixes #648

### DIFF
--- a/src/Format/Tcformat.c
+++ b/src/Format/Tcformat.c
@@ -7918,7 +7918,8 @@ BOOL CALLBACK MainDialogProc (HWND hwndDlg, UINT uMsg, WPARAM wParam, LPARAM lPa
 
 					NormalCursor ();
 
-					ShowWaitDialog (hwndDlg, TRUE, ResumeInPlaceEncWaitThreadProc, NULL);
+					if (!Silent)
+						ShowWaitDialog (hwndDlg, TRUE, ResumeInPlaceEncWaitThreadProc, NULL);
 
 					WaitCursor();
 


idrassi:
Thank you for the contribution but there is an issue with your commit.
The first thing is that `ResumeInPlaceEncWaitThreadProc` ends up not being called at all when `Silent` is `true`, which is a problem since in this part of the code we need to scan for partitions where encryption was interrupted. A better code would be:

```
	if (!Silent)
		ShowWaitDialog (hwndDlg, TRUE, ResumeInPlaceEncWaitThreadProc, NULL);
	else
		ResumeInPlaceEncWaitThreadProc (NULL, hwndDlg);
```

The second issue is that this code is not called when /create is specified in the command line so this pull request is not really linked to issue #648 .  Am I missing something? Maybe you want the waiting dialog not to be displayed which is different from the purpose of `/silent` switch (/silent was implement to allow disabling all UI in certain operations).

itsho:
My purpose was to fix #648 
if this PR is not fixing it - than it's clearly my mistake - I'll keep looking for a way to solve it.

Thanks!

idrassi:
Thank you for finding this issue.
I checked the code and the problem is in the function TCFormatVolume in file Format.c : at line, 771, we call MountVolume without setting correctly the quiet parameter and so we end up calling MountVolume with quiet set to FALSE and that's why the waiting dialog is always displayed.

So, line 771 of Format.c should be changed as follows:

`	if (MountVolume (volParams->hwndDlg, driveNo, volParams->volumePath, volParams->password, volParams->pkcs5, volParams->pim, FALSE, FALSE, FALSE, TRUE, &mountOptions, Silent, TRUE) < 1)`

Can you please create a new pull request for this fix? I can do it but since it is you who found the issue and you already started working on the fix, it is better that you get credit for it/

PR #650, opened 2020-06-29
Update VeraCrypt branch

alishir:
Updated according to this response: https://github.com/veracrypt/VeraCrypt-DCS/issues/1#issuecomment-272230174

### DIFF
--- a/src/Boot/EFI/Readme.txt
+++ b/src/Boot/EFI/Readme.txt
@@ -8,7 +8,7 @@ Here the steps to build VeraCrypt-DCS (Visual Studio 2010 SP1 should be installe
   * Clone EDK: git clone https://github.com/tianocore/tianocore.github.io.git edk2
   * Switch to UDK2015 branche: git checkout UDK2015
   * Clone VeraCrypt-DCS as DcsPkg inside edk2 folder: git clone https://github.com/veracrypt/VeraCrypt-DCS.git DcsPkg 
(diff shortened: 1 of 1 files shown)


src/Boot/EFI/Readme.txt
```diff
@@ -8,10 +8,10 @@ Here the steps to build VeraCrypt-DCS (Visual Studio 2010 SP1 should be installe
   * Clone EDK: git clone https://github.com/tianocore/tianocore.github.io.git edk2
   * Switch to UDK2015 branche: git checkout UDK2015
   * Clone VeraCrypt-DCS as DcsPkg inside edk2 folder: git clone https://github.com/veracrypt/VeraCrypt-DCS.git DcsPkg 
-  * Switch to VeraCrypt_1.18 branche: git checkout VeraCrypt_1.18
+  * Switch to VeraCrypt_1.19 branche: git checkout VeraCrypt_1.19
```
idrassi, line 11: You should use the newest tag for the git checkout (`VeraCrypt_1.24-Update2`) because `VeraCrypt_1.19` is outdated. 
```suggestion
  * Switch to VeraCrypt_1.24-Update2 branche: git checkout VeraCrypt_1.24-Update2
```

alishir: Thanks for your comment,
But I couldn't build VeraCrypt_1.24 with edk2 UDK2015 branch.
It seems I should change edk2 branch, but unfortunately I don't know which branch in compatible with latest veracrypt.

PR #666, opened 2020-08-03
Windows Bootloader: pass correct compressed data size to decompressor

xiaolangde:
Because the ds segment register is modified when setting up the decompressor segment, loader length should be read from the cs segment.


idrassi:
Thank you for finding and fixing this issue.

PR #708, opened 2020-12-09
new xml Tag for en and de

T-Bonhagen:
Hi,
like we decide here the first pull request for add linux multilanguage support

cu

### DIFF
--- a/src/Main/Main.make
+++ b/src/Main/Main.make
@@ -250,6 +250,9 @@ prepare: $(APPNAME)
 	chmod +x $(BASE_DIR)/Setup/Linux/usr/bin/$(APPNAME)-uninstall.sh
 	cp $(BASE_DIR)/License.txt $(BASE_DIR)/Setup/Linux/usr/share/doc/$(APPNAME)/License.txt
 	cp $(BASE_DIR)/../doc/html/* "$(BASE_DIR)/Setup/Linux/usr/share/doc/$(APPNAME)/HTML"
+	mkdir -p $(BASE_DIR)/Setup/Linux/usr/share/veracrypt/languages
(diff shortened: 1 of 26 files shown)


src/Main/Main.make
```diff
@@ -250,6 +250,9 @@ prepare: $(APPNAME)
 	chmod +x $(BASE_DIR)/Setup/Linux/usr/bin/$(APPNAME)-uninstall.sh
 	cp $(BASE_DIR)/License.txt $(BASE_DIR)/Setup/Linux/usr/share/doc/$(APPNAME)/License.txt
 	cp $(BASE_DIR)/../doc/html/* "$(BASE_DIR)/Setup/Linux/usr/share/doc/$(APPNAME)/HTML"
+	mkdir -p $(BASE_DIR)/Setup/Linux/usr/share/veracrypt/languages
+	cp -r $(BASE_DIR)/../Translations/* $(BASE_DIR)/Setup/Linux/usr/share/veracrypt/languages/
+	cp $(BASE_DIR)/Common/Language.xml $(BASE_DIR)/Setup/Linux/usr/share/veracrypt/languages/Language.en.xml
```
idrassi, line 255: There no need to copy Language.en.xml since it is already embedded in the code and it will be used as fallback automatically.

idrassi (changes requested):
Can you please put all newly added entries in XML files at the end of the file?

Many translations are managed by Transifex and adding new entries in the middle changes the order of entries and it makes difficult to sync with XML files of other languages.

idrassi:
Can you please address my comments?

src/Main/Main.make
T-Bonhagen: Both done

idrassi (approved):

PR #717, opened 2020-12-24
Optimized the Makefile and added some useful documentation at the top of it.

archfrog:
This tiny, minuscule *first* contribution to VeraCrypt basically speeds up the build on modern "many-core" Un*x machines.

Initially, the removal of the "-j 4" option in the Makefile, in the build target `wxbuild`, caused my Gnome desktop to crash and log me out.  However, after much experimentation, I realized that there must be some sort of bug in Gnome that kicks in when the CPU load is 100 percent without a nice value: Suddenly the memory load grows to insane heights (16 GB on my system), causing the Gnome desktop environment to crash and log me out, after having closed all my processes.

I have tested the modifications over and over, with clean boots and with and without the `ld.gold` linker.  It appears to work well.  I used `htop` to manually monitor the memory usage and it never went above 4.2 GB of RAM on my 16 GB system.

Feel free to disregard this pull request if unacceptable, I am only making it so that everyone may build VeraCrypt a tad faster.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -9,6 +9,16 @@
 # contained in the file License.txt included in VeraCrypt binary and source
 # code distribution packages.
 #
+# For best build performance, invoke Make using this command (using GNU Gold):
+#
+#    LD=ld.gold nice make -j`nproc` ...
+#
+# Or, simply (if you don't have the GNU Gold linker on your system):
+#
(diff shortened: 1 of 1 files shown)


archfrog:
I'll redo this in another way that doesn't require reading the documentation in the `Makefile` before using it so I'm closing it.

PR #731, opened 2021-01-19
fixed typos

freddii:


multiplied
```diff
@@ -0,0 +1 @@
+src/Common/GfMul.c:253:/* element in GHASH is multiplied by x^8, this function will return  */
```
alt3r-3go, line 1: This file is some artifact of your typo search process, I presume - please remove it from the PR.

alt3r-3go (changes requested):
There's one unnecessary file added (see line comment), the rest LGTM. These are comments and sometimes in a borrowed code, so may not be meaningful to edit them to avoid polluting the git history. However from purely the spelling correction prospective this PR looks ok and @idrassi can decide whether to merge all/some of it.

alt3r-3go:
@freddii, any reason you closed the pull request? Just to make it clear - I think it's worthwhile, just maybe not all changes proposed should be taken in, because they're in the borrowed code.

freddii:
1) I don't know which are the borrowed code files.
2) i have deleted my fork of it. And i don't know how to push to this pull request again to delete the "multiplied"-file.

alt3r-3go:
I see. I think (1) could be handled by the maintainer by just picking a subset of your changes and indeed (2) makes it a bit trickier, but I think at least the PR itself is associated with the VC repo, so if needed it can be pulled from.

idrassi:
I agree with @alt3r-3go that these changes are worthwhile after droping files taken from borrowed code (like t1ha.h and sha512_sse4_x64.asm).

Since the fork was deleted I could find a way to modify or merge this pull-request. If someone has an idea, let me know. Otherwise, I can extract the changes manually and then apply them after giving credit in the commit message.

freddii:
do as you like.

PR #747, opened 2021-02-23
update Simplified Chinese translation

zwj4031:


zwj4031:
update Simplified Chinese translation

PR #751, opened 2021-03-07
Update Language.de.xml

sfrings:
Translation of buttons for moving favorites was mixed up


andreas-becker (approved):
looks comprehensible to me

Emanfawaz (approved):

PR #777, opened 2021-04-19
Make system devices work under FreeBSD

MrLightningBolt:
We query the kern.geom.conftxt sysctl for the GEOM configuration to find
the partition offset. Technically speaking it would probably be better
to link against libgeom but this is less overall intrusive. Also
includes a small fix to find the parent device of an encrypted partition
when it is a GPT partition rather than a BSD slice.


idrassi:
Thank you for the contribution. The code logic could be enhance but it is doing the job. We can enhance it further in the future. I will merge it.

PR #779, opened 2021-04-21
Add support for OpenBSD

kokokodak:
Hi, I've created patch based on jcs code (https://github.com/jcs/openbsd-ports/tree/master/security/veracrypt), with some minor changes / improvements and I'd like to make that code go upstream. Could somebody please review whether there is anything more needed? I've tested the code, it compiles without problems and the file based encryption (with FAT) works without any problem.

### DIFF
--- a/src/Build/Include/Makefile.inc
+++ b/src/Build/Include/Makefile.inc
@@ -51,7 +51,11 @@ endif
 
 
 # Embedded files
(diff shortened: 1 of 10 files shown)


idrassi:
Thank you for this contribution. The code seems OK to me. I will merge it.

PR #793, opened 2021-05-25
Make compatble with pam_tmpdir

chasonr:
pam_tmpdir creates a separate temporary directory for each user ID, and that directory is not accessible to other users. This adds some security, but conflicts with VeraCrypt. VeraCrypt ends up mounting its control file system under /tmp/users/0, and the non-root user cannot access it. The user interface then does not recognize mounted VeraCrypt volumes, and cannot dismount them.

This patch recognizes when pam_tmpdir is in use, and substitutes the temporary directory for the user that invoked sudo. The user interface can then dismount normally.

### DIFF
--- a/src/Core/Unix/CoreUnix.cpp
+++ b/src/Core/Unix/CoreUnix.cpp
@@ -24,6 +24,11 @@
 
 namespace VeraCrypt
 {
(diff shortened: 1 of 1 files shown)


idrassi:
Thank you for the fix.

PR #803, opened 2021-06-23
Upload VcFixBoot

Aholicknight:
Vcfixboot is a program used to repair the bootloader in case windows removes it or when the computer updates.
dopus_TPsJ3ownfZ
This program was originally shared on [SourceForge](https://sourceforge.net/projects/veracrypt/files/Contributions/) so I will be re-uploading the same file here in case anybody wants it. 
MD5 HASH OF VcFixBoot.exe: `A3663B4ADEB55ADF83684EDAC9902F50`
MD5 HASH OF VcFixBoot64.exe: `BA906C79E637A0D49B91E0C98B735338`


alt3r-3go:
I wonder, why would you want it in the VC source repo? I'd say it's good where it is now (contributions section in the VC downloads on SF).

Aholicknight:
@alt3r-3go Your right, I only merged this as it was on the SF and I thought this would be good to have it on Github as well.

PR #811, opened 2021-08-01
Add compatibility with GNU GCC 11 Toolchain

### LINKED ISSUE #802: Issue trying to compile Arch Linux
Hi everyone, I am an Arch user and I'm trying to compile the github version of veracrypt, but I have issues when I do the first "make".

When it gets to the Cipher.cpp file to compile, there's a lot of errors (seems linked to declarations ?) but I can't find if anyone else has the same issue. I tried Update 7 and Update 4, but same result and errors.

I save the error prompt in a file joined (Sorry for the french in the file, I don't know how to get errors in English with make on a French computer).

Can someone else me to find why I can't compile ? I don't know where to start to resolve this issue.

Thanks
[log_compile.txt](https://github.com/veracrypt/VeraCrypt/files/6692396/log_compile.txt)

viciousvex:
VeraCrypt does not compile on systems with GNU GCC 11 and higher, because parameter -std=gnu++17 is used by default with these versions. Resolves #802

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -174,6 +174,15 @@ ifeq "$(shell uname -s)" "Linux"
 
 	PLATFORM := Linux
 	C_CXX_FLAGS += -DTC_UNIX -DTC_LINUX
+
+	# GNU GCC version 11 and higher compile with -std=gnu++17 by default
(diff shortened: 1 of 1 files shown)


idrassi:
Thank you for the fix!

PR #815, opened 2021-08-11
Add toggleable support for application indicators, off by default.

unit193:
I have no idea if this is actually something you want to optionally support upstream, but I guess just close this out if not.

This tries to figure out, via wx-config, if you're using GTK2 or GTK3 and uses the associated Ayatana library.

This just adds support, off by default, for building against the Ayatana libraries for indicator support.  As I understand it, for GNOME this is somewhat required to have a tray icon as they don't support the xembed ones at all.  This of course works in Xfce too and in places that don't have indicator support, it should fall back to tray icon.  Like for tint2.

I have been led to believe tray icons aren't exactly going to work out in a wayland world at all, but as I use Xfce I mainly ignore wayland at this stage.


idrassi:
Thank you for this contribution. This is actually something that I wanted to add to VeraCrypt but I never had a chance to look into it, so I will definitely merge this PR!

PR #830, opened 2021-09-22
Update German translation

Atalanttore:
Translated newly added string.


andreas-becker (approved):
Looks reasonable to me

PR #863, opened 2021-12-17
Introduction of a more sophisticated user interface

felixreichmann:
Based on the results discussed at [https://sourceforge.net/p/veracrypt/discussion/features/thread/0158a0194f/?limit=25#5913](https://sourceforge.net/p/veracrypt/discussion/features/thread/0158a0194f/?limit=25#5913) we developed a more sophisticated user interface for VeraCrypt.
The goal of this revision is to make the benefits of VeraCrypt available to a wide group of users. By incorporating more secure defaults, it was possible to reduce the number of pages in the full-disk encryption setup wizard. However, an advanced button still allows customization of all settings, so flexibility is maintained. Furthermore, the descriptions have been revised and a checklist of recommended activities before starting encryption has been introduced. This is particularly intended to ensure that the effects of the individual options are more tangible for end users.

Attached you can find a preview of the overworked interface:
[VeraCryptWizard.pdf](https://github.com/veracrypt/VeraCrypt/files/7736381/VeraCryptWizard.pdf)


idrassi:
This PR has been merged into the branch [NewSysEncWizard](https://github.com/veracrypt/VeraCrypt/tree/NewSysEncWizard) pending tests.
@felixreichmann: There is one strange point in `BrowseFile` function in [Dlgcode.c](https://github.com/veracrypt/VeraCrypt/blob/NewSysEncWizard/src/Common/Dlgcode.c#L5089): it seem to be useless function since it doesn't do anything, the field `ofn.lpstrFile` is set to NULL and it is called for clicking browse button of Rescue Disk but nothing is expected from it.
Basically, the user cannot change where Rescue Disk is stored and the browse button is dummy. 
Is this the expected behavior or just the implementation is incomplete?

I have prepared a build based on this branch and it can be downloaded from https://veracrypt.fr/downloads/VeraCrypt_Setup_1.25.8_NewSysEncUI.zip (official signature + PGP signature).

idrassi:
Addition: there is a bug in PIM handling because if I check "Modify PIM" and set a custom PIN, then pretest fail because the correct PIM is rejected. It seems as if the new code is not using the modified PIM but I didn't debug to analyze the problem.

idrassi:
Addition 2: I have found a buffer overrun bug in the new code and I fixed it (https://github.com/veracrypt/VeraCrypt/commit/70e9868c8e4f4af6345a50d153f82803e6a09bbc). This doesn't solve the PIM bug (it seems the PIM is just ignored).

I have updated the [above link](https://veracrypt.fr/downloads/VeraCrypt_Setup_1.25.8_NewSysEncUI.zip) to include the new fix.

Van-Fouran:
@idrassi Thank you for your review. I will have a look at the remarks and come back to you.

felixreichmann:
@idrassi Thank you for your review. I will have a look at the remarks and come back to you.

Van-Fouran:
We fixed the the bugs and I created a new pull request with my private GitHub Account. Please see #957

felixreichmann:
Closed, since the new pull request #957 was created.

PR #913, opened 2022-03-23
Update Corsican translation for commit 5383190

Patriccollu:
Hello,

 * https://github.com/veracrypt/VeraCrypt/commit/538319051857a8fb8e9e8c4f1048ab53e9c26b40 Block upgrade of VeraCrypt is the system is encrypted using RIPEMD-160 or GOST89 since they are not supported anymore

Best regards,
Patriccollu.


idrassi:
Thank you!

PR #914, opened 2022-03-25
Update Language.zh-cn.xml

nkh0472:


idrassi:
Thank you for the update

PR #957, opened 2022-07-27
New sys enc wizard

Van-Fouran:
This pull request is a replacement for pull request #863 .
We have fixed all the bugs mentioned there. Since we can't directly edit the branch opened by @idrassi , we have created a fork of this branch and would like to merge all changes into the master now.

Explanation of the bugs:
**1. "There is one strange point in BrowseFile function in Dlgcode.c: it seem to be useless function since it doesn't do anything, the field ofn.lpstrFile is set to NULL and it is called for clicking browse button of Rescue Disk but nothing is expected from it.
Basically, the user cannot change where Rescue Disk is stored and the browse button is dummy.
Is this the expected behavior or just the implementation is incomplete?"**
According to the Microsoft Documentation it is correct that lpstrFile is initialized with NULL.
Reference: [https://docs.microsoft.com/en-us/windows/win32/api/commdlg/ns-commdlg-openfilenamew](https://docs.microsoft.com/en-us/windows/win32/api/commdlg/ns-commdlg-openfilenamew)
"The file name used to initialize the File Name edit control. The first character of this buffer must be NULL if initialization is not necessary. When the GetOpenFileName or GetSaveFileName function returns successfully, this buffer contains the drive designator, path, file name, and extension of the selected file."
Nevertheless, it was right, that the browse button was some kind of a dummy button. We fixed it by correcting a false return statement.

**2. "there is a bug in PIM handling because if I check "Modify PIM" and set a custom PIN, then pretest fail because the correct PIM is rejected. It seems as if the new code is not using the modified PIM but I didn't debug to analyze the problem."**
The master key was generated during the password page. At this time the PIM was not yet modified within the new wizzard. Thats why the default PIM value was used. We have fixed it by generating the master key one page later in case the enable PIM option was activated.

Additionally to the mentioned bugs, we fixed the following:
1. The newly needed password file, containing the 1 Mio. most used passwords was deleted after running the sign_test.bat script the first time.
2. We modified the displayed text slightly two times.
3. Removed the entry of RIPEMD-160 within the Product64.wxs file, since RIPEMD-160 is not available within VeraCrypt anymore. The corresponding error during the build process is fixed by this modification.
4. Added a german translation for the new interface.


idrassi:
Thank you @Van-Fouran for this new submission. It is getting closer to 
Can you please help solve the conflicts on the files Translations/Language.de.xml and src/Common/Language.xml?

There are 2 rules to follow for XML language files:
- Existing lines should not never be removed even if they are not used anymore
- New entries should always be added at the end of the XML files.

The reason behind these rules is that we don't want to change the ordering of the entries which should remain the same across all languages. 

Other than this, the main point that we need to check is the automatic translation of the password to US keyboard. It looks risky and it is possible that it will create problems for non-western users. Also, it seems to change the existing behavior of VeraCrypt which can create confusion for existing users. Any feedback on this and the tests around it?

Van-Fouran:
**Can you please help solve the conflicts on the files Translations/Language.de.xml and src/Common/Language.xml?**
Sure, I will take care of this and will check if I followed the rules.

----

**Other than this, the main point that we need to check is the automatic translation of the password to US keyboard. It looks risky and it is possible that it will create problems for non-western users. Also, it seems to change the existing behavior of VeraCrypt which can create confusion for existing users. Any feedback on this and the tests around it?**

**Background information: Why does the keyboard layout has to be changed?**
In general there is the issue that in the pre-boot environment only the qwerty keyboard layout is support. So you have to make sure, that the user can enter the same password at the set-up of the encryption and in the pre-boot environment.
Otherwise you get the same situation as on Microsoft BitLocker: If, for example, a german user set a password including the letter "z", he has to enter a "y" within the pre-boot environment due to the different keyboard layout. This can lock computers, because users keep entering "false" passwords and are not aware of this technical issue.

**Possible Solutions:**
1. **Just post a notice**
In my perspective this is the worst solution. User may not understand what they need to do. Further, they have to manually check how the password has to be transformed. The possibility is listed for completeness, but I would exclude it from any further discussion.
2. **Change the keyboard layout during the setup process to the qwerty layout**
This is the current default behavior of VeraCrypt. It ensures that the user enters the password in the same way as it has to be entered within the pre-boot environment.
With this solution the end-user may gets a notification from the operating system that the keyboard layout has been changed. Further, if the user displays the password, it is not the expected password.
If we stick with this solution I recommend to put at least a notice for the users in the setup dialog. But I am not sure if lay users understand why they do not see the expected password based on a short notice.
3. **Transform the password in the background**
This is the solution proposed within the PR. The keyboard layout is not changed during the setup phase and the user enters the password in the current keyboard layout. Therefore, the password looks as expected when displaying it and there is no notification by the OS. After clicking on the next button the password is transformed into the qwerty layout (e.g. all german "z" are exchanges against "y"). This process is completely transparent for the end-user.

@idrassi I guess you see the risk that the process strongly relies on the quality of password transformation.
In general I can confirm this risk. As a countermeasure, we have included an abort of the process if the transformation cannot be performed as planned (for example, if the QWERTY layout has been removed from the Windows operating system). This ensures that at least the encryption process does not continue.
We have successfully tested the process with the following keyboard layouts: QWERTZ (German), QWERTZ (Croatia) and QWERTY
If wished, I am happy to run tests for other layouts as well.

Van-Fouran:
@idrassi The merge conflicts within the language files are fixed.

idrassi:
Thank you @Van-Fouran for the explanation about the password transformation and for fixing the merge conflicts.

I'm still not comfortable with this password transformation approach. For example, with this change, the password displayed in the UI of the encryption wizard and the password displayed in the bootloader using F5 key will look different and this will cause issues for users.

Your idea is based on the will to make users think that what they type is what ends up being used even-though this is not true. I'm not sure I understand the benefit of this.

One parameter I also take into account is troubleshooting boot issues and data recovery. With the current VeraCrypt approach, the system encryption password can be copied to a password manager and it can be later be used as it is when doing data recovery or booting on another machine. With the new approach, the value stored in the password manager is of no use since it is different from the real password.

To summarize:

- With current VeraCrypt approach:
   - The password displayed in wizard UI is the same as the one displayed in bootloader with F5.
   - The password displayed is the one that will be used by VeraCrypt for actual encryption.
   - The password can be stored in password manager for later use during troubleshooting or data recovery.

- With the new approach:
   - The password displayed in wizard UI is different from the one displayed in bootloader with F5.
   - The password displayed is different from the one used by VeraCrypt for actual encryption
   - There are potential issues with non-western keyboard layouts
   - The password entered in the UI cannot be saved for later use since it will be different from real password.

As you can see, there are so many side effects and implications of this new approach and that's why I'm reluctant to rollout this as the default for VeraCrypt users.

The only case that I see where this approach can be useful if when the PC doesn't have US-keyboard layout installed (for example, user manually removed it and only left its local keyboard layout). In this case, the wizard will abort because it will not be able to switch to the US layout. As an alternative, we can propose to use the automatic password transformation instead of just aborting while warning about possible caveats for untested keyboards.

Van-Fouran:
@idrassi Thank you for your reply.
If you are fine with this I would proceed with the following approach: 
I revert the changes we have made regarding the password transformation for now. This would enable us the merge this pull request.
I have a list of topics that could be reviewed in the future. I would include the keyboard aspect here so that it could be looked at more closely in the future.

Van-Fouran:
Today's usual behavior in handling the keyboard layout is restored.
I have added the necessary lines of code again. The new functions we created are no longer called. I left them currently still in the code, so that one could fall back on it if necessary. If the final decision is made not to pursue this option any further, the artifacts can be removed.

idrassi:
Thank you for the changes.
Having dead code is not a good thing especially when others are doing code audits. We should be doing code cleanup. The option to automatically translate password makes sens when there is no US keyboard layout as I said previously so we can keep the code for future changes but it is better to put it in a separate file. This will help others audit this new changes and ensure that it is safe.
Thanks.

Meanwhile, I will create a build from this PR to seek comments from some users.

Van-Fouran:
> Thank you for the changes.
> 
> Having dead code is not a good thing especially when others are doing code audits. We should be doing code cleanup. The option to automatically translate password makes sens when there is no US keyboard layout as I said previously so we can keep the code for future changes but it is better to put it in a separate file. This will help others audit this new changes and ensure that it is safe.
> 
> Thanks.
> 
> 
> 
> Meanwhile, I will create a build from this PR to seek comments from some users.

Ok, thanks. I will clean the code.

Van-Fouran:
Done.

idrassi:
Sorry for come back again to language XML files but there is still a small ordering issue:

- in Language.xml, the new entries were added after the last item "SYS_ENCRYPTION_UPGRADE_UNSUPPORTED_ALGORITHM" instead of being after
- in Language.de.xml, the ordering of the items not the same as Language.xml (you can see this in the diff of the PR).

I know this annoying but the ordering the important as I explained above. For example, Transifex uses the ordering of the items in the XML files to match the original English entry with its corresponding translation. 
Can you please handle this last point?

My idea is to share a beta build with these changed on Sourceforge to seek feedback from long time community users. I expected many comments and maybe criticism and bugs so the coming weeks will certainly be intense. We will see how things go.

Van-Fouran:
@idrassi I fixed it.

idrassi:
Unfortunately there is still the issue in Language.de.xml so maybe my previous comment was not understood fully.
Here are the issues:

- "`SYSENC_TYPE_PAGE_TITLE`" position changed: it was after "`SYSENC_PRE_DRIVE_ANALYSIS_HELP`" but now it is after "`SYSENC_MULTI_BOOT_ADJACENT_SYS_TITLE`"
- "`SYS_ENCRYPTION_PRETEST_RESULT_TITLE`" position changed: it was after "`SYS_ENCRYPTION_PRETEST_INFO2_PORTION_4`" but now it is after "`RESCUE_DISK_RECORDING_TITLE`"
- "`RESCUE_DISK_EFI_INFO`" position changed: it was after "`PRE-BOOT`" but now it is after "`ITERATIONS`". 

Below are screenshots illustrating the first issue as taken from Github diff of the PR.

it is not difficult to fix after merging the PR but I prefer that PR be complete before merge so that the owner of the changes is easily identified.
Thank you for your efforts.

diff1
diff2

Van-Fouran:
> Unfortunately there is still the issue in Language.de.xml so maybe my previous comment was not understood fully. Here are the issues:
> 
>     * "`SYSENC_TYPE_PAGE_TITLE`" position changed: it was after "`SYSENC_PRE_DRIVE_ANALYSIS_HELP`" but now it is after "`SYSENC_MULTI_BOOT_ADJACENT_SYS_TITLE`"
> 
>     * "`SYS_ENCRYPTION_PRETEST_RESULT_TITLE`" position changed: it was after "`SYS_ENCRYPTION_PRETEST_INFO2_PORTION_4`" but now it is after "`RESCUE_DISK_RECORDING_TITLE`"
> 
>     * "`RESCUE_DISK_EFI_INFO`" position changed: it was after "`PRE-BOOT`" but now it is after "`ITERATIONS`".
> 
> 
> Below are screenshots illustrating the first issue as taken from Github diff of the PR.
> 
> it is not difficult to fix after merging the PR but I prefer that PR be complete before merge so that the owner of the changes is easily identified. Thank you for your efforts.
> 
> diff1 diff2

Fixed.

idrassi:
Hi @Van-Fouran,
I created a build after merging this PR. There was an issue with CHM documentation because the div technique to hide/show newly added section doesn't work and so I have to manually disable it to make CHM documentation useful. There is still missing link in CHM left side tree but I will handle this later.
The build is available for now on the following links:

- https://veracrypt.fr/downloads/VeraCrypt%20Portable%201.26.1.exe (https://veracrypt.fr/downloads/VeraCrypt%20Portable%201.26.1.exe.sig)
- https://veracrypt.fr/downloads/VeraCrypt%20Setup%201.26.1.exe (https://veracrypt.fr/downloads/VeraCrypt%20Setup%201.26.1.exe.sig)
- https://veracrypt.fr/downloads/VeraCrypt_Setup_x64_1.26.1.msi (https://veracrypt.fr/downloads/VeraCrypt_Setup_x64_1.26.1.msi.sig)

xeropresence:
It seems like there is an issue with the builds produced with this merged.

I installed the above 1.26.1 build on a laptop running windows 11, standard qwerty keyboard layout in the English locale.
I created a strong password and proceeded to the encrypt the windows partition. I left all settings at their defaults, SHA512, AES, no changes to pim, etc.
Password was confirmed matching during the gui process, and upon reboot to the pretest, password would fail to unlock the drives. 

What did work was leaving the password field blank, and pressing enter twice to confirm blank password and no change to the pim.

I attempted to change password using the gui after booting, but using the previously created password it would fail with a mismatch, the gui does not accept a empty value so that could not be tested.

Decryption of the system drive worked fine.

Uninstalled 1.26.1 and installed 1.25.9 and repeated the same procedure, this time password prompt works as expected taking the previously created password.

Van-Fouran:
> It seems like there is an issue with the builds produced with this merged.
> 
> I installed the above 1.26.1 build on a laptop running windows 11, standard qwerty keyboard layout in the English locale. I created a strong password and proceeded to encrypt the windows partition. I left all settings at their defaults, SHA512, AES, no changes to pim, etc. Password was confirmed matching during the gui process, and upon reboot to the pretest, password would fail to unlock the partition.
> 
> What did work was leaving the password field blank, and pressing enter twice to confirm blank password and no change to the pim.
> 
> I assumed this might be a bug with the pretest, so I let the disk encryption run to completion, rebooted after and same results occurred.
> 
> I attempted to change password using the gui after booting, but using the previously created password it would fail with a mismatch, the gui does not accept a empty value so that could not be tested.
> 
> Decryption of the system drive worked fine.
> 
> Uninstalled 1.26.1 and installed 1.25.9 and repeated the same procedure, this time password prompt works as expected taking the previously created password.

Hello,
I tried to recreate the problem, however I did not succeed. I waited for the pretest as well as the encryption itself. Can you please check if I used a similar configuration as you?
Have you turned off Fast Startup?


**Virtual Machine Config:**
The installation was a fresh Windows 11 Pro installation only with VMWare Tools and VeraCrypt as additional applications.
image

**VeraCrypt Config**
2022-11-29 12_58_57-Windows 11 - VMware Workstation
2022-11-29 12_59_28-Windows 11 - VMware Workstation

xeropresence:
I'm pretty sure I was prompted to disable fast startup, during either the install or the initial run of VC and disabled it as prompted.


Things that were different:

Physical install
Windows 11 ver 21h2, 22000.1219
AMD 6900HS
Longer password  20> <40 chars, had spaces

Thing's I might have done differently:
I'm not sure, but I think I did change the hash algorithm to blake, others then back to the default sha512, possibly repeating that with the encryption but I think that is less likely.

xeropresence:
Oh, Secure boot should still be enabled as well.

Van-Fouran:
> I'm pretty sure I was prompted to disable fast startup, during either the install or the initial run of VC and disabled it as prompted.

Yes, you are prompted.

> 
> Things that were different:
> 
> Physical install Windows 11 Enterprise ver 21h2, 22000.1219 AMD 6900HS Longer password 20> <40 chars, had spaces
> 
> Thing's I might have done differently: I'm not sure, but I think I did change the hash algorithm to blake, others then back to the default sha512, possibly repeating that with the encryption but I think that is less likely.

I made the following changes, but on my system the encryption still works:

- I have chosen a password with 23 characters including spaces
- I have chosen BLAKE2s-256 as hash algorithm
- I have enabled Secure Boot (I know that this option caused issues in prior versions)

2022-11-30 00_01_08-Windows 11 - VMware Workstation

xeropresence:
Just to be clear, I did use the default sha512, but I had selected the different drop down items before returning to sha512.  (Maybe some issue with the gui setting a strange value?)

I can't think of anything else that could be done differently, but I cannot retest it at the moment personally. If you have anymore questions please let me know.

wendig0x:
@idrassi I ask you to completely roll back this pull request (#957), because I consider it suspicious. These changes require detailed analysis. Also, 1. the problem described by @xeropresence is seen by many users 2. сhanges to the GUI hide important settings from the user.

Van-Fouran:
Hi @wendig0x 
Thank you for your feedback. 

I agree with you that the changes have an impact on the interface of VeraCrypt and therefore should be tested. I can assure you that we have been analyzing the impact and quality of the pull request for quite some time.

If you have any further concrete suggestions, please let us look at them and make another pull request if necessary.


> 1. the problem described by @xeropresence is seen by many users 
I have tried to reconstruct the described problem to analyze it. So far, however, it has not occurred on my system. Can you please send me the references to the other users so that I can contact them and analyze the problem further?
Did you experience the problem as well?

> 2. сhanges to the GUI hide important settings from the user.
What settings do you mean here exactly?
The goal of this pull request is to make the interface of VeraCrypt more user friendly, especially for non-technical users. We didn't remove any settings, we just hid technical settings behind an Advanced button. This ensures that users without technical background do not unknowingly change secure default settings, but experienced users can still adjust all settings. This approach is also recommended in relevant research papers.

> Attention! Using version 1.26.1 is not safe!
What factors lead you to conclude that the version is unsafe?

wendig0x:
> I have tried to reconstruct the described problem to analyze it. So far, however, it has not occurred on my system. Can you please send me the references to the other users so that I can contact them and analyze the problem further?
> Did you experience the problem as well?

I and some people from my environment faced. Also, it seemed to me that there was some kind of error in the handling of characters like these: # $ % & ' ( ) * + , - . / : ; <=.

> This ensures that users without technical background do not unknowingly change secure default settings

What combinations of hash and encryption algorithm do you consider less secure?

> What factors lead you to conclude that the version is unsafe?

Why is there no collection of entropy of mouse movement with system encryption, for example?

Van-Fouran:
> > I have tried to reconstruct the described problem to analyze it. So far, however, it has not occurred on my system. Can you please send me the references to the other users so that I can contact them and analyze the problem further?
> > Did you experience the problem as well?
> 
> I and some people from my environment faced. Also, it seemed to me that there was some kind of error in the handling of characters like these: # $ % & ' ( ) * + , - . / : ; <=.

I listed my test environment here: [https://github.com/veracrypt/VeraCrypt/pull/957#issuecomment-1330539228](https://github.com/veracrypt/VeraCrypt/pull/957#issuecomment-1330539228)
Can you please specify what was different in your environment, so I can reproduce the issue?

> > This ensures that users without technical background do not unknowingly change secure default settings
> 
> What combinations of hash and encryption algorithm do you consider less secure?

The problem is already the selection. Many users are completely unfamiliar with the algorithms and are overwhelmed with them as soon as a selection is displayed. Therefore, we make a selection that we consider to be basically secure, so that the user does not have to do this anymore. However, experienced users can still customize the settings. From a usability perspective, the algorithms should not be displayed as a mandatory selection menu in the main process.

> 
> > What factors lead you to conclude that the version is unsafe?
> 
> Why is there no collection of entropy of mouse movement with system encryption, for example?

It has been found out that a part of users completely abort the encryption process at this step, as they are unable to comprehend it. We have removed it in our proposal, as we believe that this menu only marginally affects security, but has a strong impact on usability.
Feel free to make a suggestion on how to make the page more optimal.

Van-Fouran:
> I listed my test environment here: https://github.com/veracrypt/VeraCrypt/pull/957#issuecomment-1330539228
Can you please specify what was different in your environment, so I can reproduce the issue?

I would ask you to tell me under what conditions your problems occurred. Since I can't detect them in my current test environment, I can't look for a cause otherwise.

 
 

> > Many users are completely unfamiliar with the algorithms and are overwhelmed with them as soon as a selection is displayed.
> 
> Default values will be attacked first.
> 

At this point, a clear distinction must be made between settings (e.g. algorithm used) and secrets (e.g. passwords). While default values should of course be avoided for secrets, this is different for settings. Security should not be based on keeping the algorithm used obfuscate. In this respect, using secure default settings, such as preselecting an encryption or hash algorithm as default value, is not a security risk.

> > We have removed it in our proposal, as we believe that this menu only marginally affects security, but has a strong impact on usability.
> 
> Are you joking? Provide links to scientific articles that you relied on in your "research". Especially where bio-PRNG is considered "unimportant". With the same success, you can set the default password to "admin". To make it convenient for users.

Since there have been few studies on VeraCrypt itself, the interface was largely inspired by this work on its predecessor TrueCrypt: [https://dl.acm.org/doi/abs/10.1145/2435349.2435360](https://dl.acm.org/doi/abs/10.1145/2435349.2435360).

The paper examines the wizard for creating encrypted volumes, but the elements are very analogous to those from disk encryption.

To be fair, the referenced paper did not look at the "mouse movement" page, as it was not part of the scope examined.

However, I know from more recent non-public investigations that this page poses a greater challenge to users. Thus, if this page is included in its current form, there is a risk that users will avoid using VeraCrypt altogether. In my view, that should not be the goal. Rather, we should adapt the VeraCrypt interface so that it can be used by end users.

The comparison regarding the admin password is incorrect. Entropy is still used in the generation of key material by pseudo-random generators. However, these do not **additionally** use the mouse movements as source.
By the way, if you compare how other established full disk encryption solutions generate the entropy, you will notice that none uses mouse movement during the setup process as entropy source.

However, if there is a way to implement entropy generation by mouse movements in a way that is usable for the end user, there is nothing wrong with it.
Have you any suggestions on how to incorporate the process into the VeraCrypt Wizard so that it remains usable?

wendig0x:
> Security should not be based on keeping the algorithm used obfuscate.

You don't understand what these words mean.  
I recommend checking out: https://blog.elcomsoft.com/2021/06/breaking-veracrypt-obtaining-and-extracting-on-the-fly-encryption-keys/

@In VeraCrypt, information about the encryption algorithm or the KDF is never saved in the disk header. When encrypting the disk, users have the choice of 15 encryption algorithms (including combinations) and 5 hash functions. That makes it 15×5=75 possible combinations of symmetric ciphers and one-way hash functions to try. If you don't know exactly which cipher and which hash function has been used to encrypt the container, you'll have to try all of the 75 combinations during the attack starting with the default combination of AES-256 encryption and SHA-512 hash function.@

> I would ask you to tell me under what conditions your problems occurred. Since I can't detect them in my current test environment, I can't look for a cause otherwise.

There is no point in discussing, because your pull request should be cancelled.

I will not continue the discussion, because you do not understand cryptanalysis at all.

idrassi:
I’m back after a prolonged absence.
@xeropresence @wendig0x @Van-Fouran 
This PR will be reverted on master and then put on a dedicated branch.
Version 1.26.1 is a test version whose objective was to gather feedback from the community about ideas proposed by @Van-Fouran.
Clearly there are various issues that need to be addressed before putting parts of these changes into an official release. There is also a gap of security expectations and security target between current VeraCrypt approach and the new approach. This is a very important point that should not be ignored.

PR #963, opened 2022-08-29
Added line breaks between new copyright holders

sandakersmann:
Added line breaks between new copyright holders


idrassi:
Thanks for the contribution.

PR #999, opened 2022-11-14
wxWidget assert failure bug

An7ar35:
Removal of any of the any of `wxALIGN_RIGHT`/`wxALIGN_CENTRE_HORIZONT…AL`/`wxALIGN_BOTTOM`/`wxALIGN_CENTRE_VERTICAL` when used in conjunction with `wxEXPAND`

wxWidget has an "assert" that gets triggered anytime `wxEXPAND` gets used with an alignment property (a now useless override). This leads to a popup backtrace every time a non-compliant widget is opened.

### DIFF
--- a/src/Main/Forms/Forms.cpp
+++ b/src/Main/Forms/Forms.cpp
@@ -586,24 +586,24 @@ WizardFrameBase::WizardFrameBase( wxWindow* parent, wxWindowID id, const wxStrin
 	bSizer70->Add( 0, 0, 1, wxEXPAND, 5 );
 	
 	HelpButton = new wxButton( MainPanel, wxID_HELP, _("IDHELP"), wxDefaultPosition, wxDefaultSize, 0 );
-	bSizer70->Add( HelpButton, 0, wxALL|wxALIGN_RIGHT|wxALIGN_CENTER_VERTICAL, 5 );
+	bSizer70->Add( HelpButton, 0, wxALL|wxALIGN_CENTER_VERTICAL, 5 );
 	
 	
-	bSizer70->Add( 0, 0, 0, wxLEFT|wxALIGN_RIGHT, 5 );
+	bSizer70->Add( 2, 0, 0, wxLEFT, 5 );
 	
 	PreviousButton = new wxButton( MainPanel, wxID_ANY, _("PREV"), wxDefaultPosition, wxDefaultSize, 0 );
-	bSizer70->Add( PreviousButton, 0, wxTOP|wxBOTTOM|wxLEFT|wxALIGN_RIGHT|wxALIGN_CENTER_VERTICAL, 5 );
+	bSizer70->Add( PreviousButton, 0, wxTOP|wxBOTTOM|wxLEFT|wxALIGN_CENTER_VERTICAL, 5 );
 	
 	NextButton = new wxButton( MainPanel, wxID_ANY, _("NEXT"), wxDefaultPosition, wxDefaultSize, 0|wxWANTS_CHARS );
 	NextButton->SetDefault();
 	
-	bSizer70->Add( NextButton, 0, wxTOP|wxBOTTOM|wxRIGHT|wxALIGN_RIGHT|wxALIGN_CENTER_VERTICAL, 5 );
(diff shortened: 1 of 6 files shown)


An7ar35:
It should be noted that this issue was found on Linux using wxWidget 3.2.1.

bugtracker2019:
Hi,

this issue also interests me a lot.
Just curious: with your changes in the three files, does Veracrypt "behave" correctly now (no pop-up windows anymore)? Have you tested all possibilities?
If I understand correctly one has to remove ANY lines containing wxALIGN... AND wxEXPAND, right?
Just checked Forms.cpp (`awk '/wxALIGN/ && /wxEXPAND/' Forms.cpp`) and there are still a bunch of lines left, you haven't worked out. Do I miss sth. here?

bugtracker2019:
[This](https://github.com/archlinux/svntogit-community/blob/packages/veracrypt/trunk/wx-3.2-size-warnings.patch) would fix it, wouldn't it?

bugtracker2019:
Hi An7ar35,

although freswa's fix from the archlinux-git get's rid of the annoying popup-window, it's rather quick and dirty. I'm sharing your philosophy and would like to see things done as recommended (aka "right").

I would like to support you in this matter, but my programming skills are on a VERY basic level.^^
Maybe you can point me in the right direction, i.e. do you already have sorted out which other combinations of wxALIGN* take effect? Maybe there are more helpful infos on wxwidgets.org, where the issue was also [mentioned this summer](https://www.wxwidgets.org/blog/2021/06/sizer-asserts-in-wxwidgets-3.1/).

Slightly off-topic: which IDE do you use to find the corresponding lines with wxALIGN* & wxEXPAND? If it's geany what's your input in the search mask?

An7ar35:
Hi @bugtracker2019 

I started last night with the `wxALIGN_RIGHT`/`wxALIGN_LEFT` inside `wxHORIZONTAL` sizers issue but it looks like there were only 2 (e2ba714 and 05284de). So I think we are actually all good now :crossed_fingers: .

Although, it wouldn't hurt double checking all the UI panels manually to make sure there are no more errors. I didn't get any when I went through the UI but you never know! If you want to have a crack at it and do some checking on your side that would be appreciated. 2 sets of eyeballs are better than 1.

If you are on Linux and you run into git/compilation problems, hit me up.

P.s.: for your last question; I use [CLion](https://www.jetbrains.com/clion/). I used the catch-all "find in files" functionality from the source dir and looked for any things like `(wxALIGN_[A-Z_]*[A-Za-z\ \|]*wxEXPAND)|(wxEXPAND[A-Za-z\ \|]*wxALIGN_[A-Z_]*)` for example. Once thing that bears mention is that `wxTextCtrl` items don't trigger the assert failure thus I left them alone.

An7ar35:
I should probably add for prosperity that the `src/Main/Forms/TrueCrypt.fbp` file is the [wxFormBuilder](https://github.com/wxFormBuilder/wxFormBuilder) template which generates the `src/Main/Forms/Forms.h/cpp` files normally.

In the future, changing the UI should really be done via [wxFormBuilder](https://github.com/wxFormBuilder/wxFormBuilder) and the code re-generated to be properly compliant. Commit be34e4b deals with the breakage in that scenario so it should be fine from now on to use that workflow.

bugtracker2019:
Hey @An7ar35,

just a short feedback: I compiled Veracrypt with the five patched files and it looks good so far. :+1: 
If you're interested, on [pastebin](https://pastebin.com/r5pJagVZ) is my compiling-output.

An7ar35:
Hi @bugtracker2019,

Awesome! Thank you.

I got the same output. Since the compile warning for `wxFileName::Normalize()` touches on the UI/wxWidget scope I've sorted that out (0314044). The others are beyond this PR I think so I'm leaving them alone for the time being.

bugtracker2019:
Nice work. Let's hope your changes will make it into the next version! :+1:

idrassi:
Thank you @An7ar35 for this PR, it is very helpful!
In parallel to this fix, I have pushed a commit that silences the wxWidgets sizer flags assert using the `WXSUPPRESS_SIZER_FLAGS_CHECK` environment variable: https://github.com/veracrypt/VeraCrypt/commit/8e443c25cce0edf51335019cbbab8ae11fa7df21

idrassi (approved):

Issue #968, opened 2022-09-06
vnconfig option "-c" does not exist on OpenBSD


nochnaccount:
Hello,

Running veracrypt on OpenBSD-current via:
```
$ veracrypt -t -k "" --pim=0 --protect-hidden=no /home/media/vc /mnt/veracrypt1
```
gives the following error:
```
Enter password for /home/media/vc:
Error: vnconfig: unknown option -- c
usage: vnconfig [-v] [-k | -K rounds [-S saltfile]] [-t disktype]
[vnd_dev] image
       vnconfig -l [vnd_dev]
              vnconfig -u [-v] vnd_dev
```
That '-c' may have sneaked in from FreeBSD's 4.xx vnconfig as far
as I can tell.
```
$ head Core/Unix/OpenBSD/CoreOpenBSD.cpp 

/* $OpenBSD$ */
/*
 Based on FreeBSD/CoreFreeBSD.cpp

 Derived from source code of TrueCrypt 7.1a, which is
 Copyright (c) 2008-2012 TrueCrypt Developers Association and which is governed
 by the TrueCrypt License 3.0.

 Modifications and additions to the original source code (contained in this file)
 and all other portions of this file are Copyright (c) 2013-2017 IDRIX
```
The following patch makes it work again here on OpenBSD.
```
--- Core/Unix/OpenBSD/CoreOpenBSD.cpp.dist	Tue Aug 30 11:32:39 2022
+++ Core/Unix/OpenBSD/CoreOpenBSD.cpp	Tue Aug 30 11:34:49 2022
@@ -69,7 +69,7 @@
 		if (freeVnd == -1)
 			throw "couldn't find free vnd";
 
-		args.push_back ("-c");
+		// args.push_back ("-c");
 
 		stringstream freePath;
 		freePath << "vnd" << freeVnd;

```

idrassi:
Thank you for reporting this issue.
PR https://github.com/veracrypt/VeraCrypt/pull/1002 which fixes this issue has been merged into master.

PR #1002, opened 2022-11-23
OpenBSD vnconfig no longer accepts -c option

ZhanYF:
vnconfig used to accept a `-c` option for configuring virtual disk node, but this option is not required as configuring virtual disk node is always the default operation. Since OpenBSD 5.6, this option is no longer accepted and will result in an error, preventing veracrypt from mounting any volume.

This change removes the `-c` option when invoking vnconfig.

Tested on OpenBSD 7.2, same patch also tested in https://github.com/veracrypt/VeraCrypt/issues/968

### DIFF
--- a/src/Core/Unix/OpenBSD/CoreOpenBSD.cpp
+++ b/src/Core/Unix/OpenBSD/CoreOpenBSD.cpp
@@ -69,8 +69,6 @@ namespace VeraCrypt
 		if (freeVnd == -1)
 			throw "couldn't find free vnd";
 
-		args.push_back ("-c");
-
 		stringstream freePath;
 		freePath << "vnd" << freeVnd;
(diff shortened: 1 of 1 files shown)


kokokodak:
Hi, I've tested this patch on OpenBSD 7.2 and current. It's also currently being used in veracrypt port (https://github.com/openbsd/ports/blob/master/security/veracrypt/patches/patch-Core_Unix_OpenBSD_CoreOpenBSD_cpp) and the affected file is OpenBSD-only.

It's true that the -c option has been removed between 6.5 and 6.6 (in particular this commit https://github.com/openbsd/src/commit/7d9a3feb9d7f07a70b116662860d608742e23a97)

It works fine and I didn't find any issues with it.

idrassi:
Thank you for the fix.

PR #1025, opened 2023-01-13
Fix error when checking for free space

bogdro:
When creating a totally new volume on the command line (Linux), VeraCrypt displays an error:
```
Error: Failed to get file system statistics (error 2: No such file or directory)
```
This is caused by passing an empty path to `wxGetDiskSpace`.
My fix is a simple workaround for this - if the path is empty, just don't call `wxGetDiskSpace`.

Perhaps a better solution would be to pass something like the basedir of the path if the file doesn't exist, but this is what I can do with my knowledge.

### DIFF
--- a/src/Main/TextUserInterface.cpp
+++ b/src/Main/TextUserInterface.cpp
@@ -648,7 +648,8 @@ namespace VeraCrypt
 		{
 			uint64 AvailableDiskSpace = 0;
 			wxLongLong diskSpace = 0;
-			if (wxGetDiskSpace (wxFileName (wstring (options->Path)).GetPath(), nullptr, &diskSpace))
+			if (!wxFileName (wstring (options->Path)).GetPath().IsEmpty() &&
+				wxGetDiskSpace (wxFileName (wstring (options->Path)).GetPath(), nullptr, &diskSpace))
 			{
 				AvailableDiskSpace = (uint64) diskSpace.GetValue ();
 				if (maxVolumeSize > AvailableDiskSpace)


bogdro:
Sure.
Call `--create` on the command line, enter `1` as the volume type and any non-existing path as the volume path.
Before the fix:
```
$ ./veracrypt-1.25.9 --create
Volume type:
 1) Normal
 2) Hidden
Select [1]: 1

Enter volume path: blah
07:34:37 PM: Error: Failed to get file system statistics (error 2: No such file or directory)

Enter volume size (sizeK/size[M]/sizeG.sizeT/max): ^C
```
After the fix:
```
$ ./veracrypt-1.25.9-mod --create
Volume type:
 1) Normal
 2) Hidden
Select [1]: 1

Enter volume path: blah

Enter volume size (sizeK/size[M]/sizeG.sizeT/max): ^C
```
System type:
```
$ uname -a
Linux cepheus 5.16.13-desktop-1omv4003 #1 SMP PREEMPT Wed Mar 9 19:33:24 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

bogdro:
Hi.

Thanks for the reply.

Yes, it does actually work (=no error shown) when I put `./blah` or `/home/blah`. It doesn't work when I put `/home/xxx/blah` (that's a literal "xxx", a username which doesn't exist), with the same error message. Obviously, the file wouldn't have been created, and the error would be the same as in any other "usual tool" (touch, cat, etc.), but in this case the absolute path doesn't help. This time it's a real user error, though, and not just not knowing/using the right way to write filenames.

Anyway, when I put just 'blah', an error is shown, but everything works, so the error message may be misleading or suggesting that the user is doing something wrong while in fact this is just a minor potential issue (creating a file bigger than the free space, I guess). So, another possible solution: if the path doesn't contain "/" (or maybe "\\" on Windows), prepend "./" at the beginning or if the resulting path would be empty, simply use "." (current directory) to just get the filesystem statistics, because that's where the file would be created - in the current directory.

idrassi:
I'm for the solution 1 where current directory is assumed if not path is found in the five file name.
This means that to fix this issue, we should use "." in  the call to wxGetDiskSpace if the path is empty, as @bogdro suggested. 

I can implement this change but it is better if @bogdro can implement it in this PR since he is the one behind the idea.

idrassi:
I have pushed a change that implements the idea proposed by @bogdro: https://github.com/veracrypt/VeraCrypt/commit/b872702309b20f2467d58652c3b00493ac4058bd

So I will close this PR.
Thank you again for raising this issue.

PR #1026, opened 2023-01-13
Keep PIM secret and re-ask when user entered a wrong value

bogdro:
Since PIM is a part of the user's secret (the key is derived from it), as it is written:

> The PIM is treated like a secret value that must be entered by the user each time alongside the password.

(https://veracrypt.fr/en/Personal%20Iterations%20Multiplier%20(PIM).html?), it should also be hidden from view and not echoed to the terminal.

This patch addresses this issue.

Apart from this, PIM should be allowed to be re-entered, just like the password is, in case decryption fails. Otherwise, if the user enters the correct password and an incorrect PIM, he/she doesn't have a chance to fix the mistake - the process must be re-started.

This issue is also addressed.

### DIFF
--- a/src/Main/TextUserInterface.cpp
+++ b/src/Main/TextUserInterface.cpp
@@ -175,9 +175,13 @@ namespace VeraCrypt
 		wxString msg = _("Enter new PIM: ");
 		if (!message.empty())
 			msg = message + L": ";
+		SetTerminalEcho (false);
+		finally_do ({ TextUserInterface::SetTerminalEcho (true); });
 		while (pim < 0)
 		{
 			wstring pimStr = AskString (msg);
+			ShowString (L"\n");
+
 			if (pimStr.empty())
 				pim = 0;
 			else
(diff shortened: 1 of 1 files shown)


src/Main/TextUserInterface.cpp
```diff
@@ -1290,7 +1294,7 @@ namespace VeraCrypt
 				options.Password = AskPassword (StringFormatter (_("Enter password for {0}"), wstring (*options.Path)));
 			}
 
-			if (!options.TrueCryptMode && (options.Pim < 0))
+			if (!options.TrueCryptMode /*&& (options.Pim < 0) re-ask in case of wrong value entered*/)
```
idrassi, line 1297: Actually this is not the correct way to fix the issue. Instead, `options.Pim `should be set to -1 little bit below when we catch the exceptions `PasswordIncorrect `and `PasswordException` like what we do for `options.Password`.

idrassi (changes requested):

idrassi:
I will merge this PR and do the changes that I mentionned on my comment directly.
Thank you.

bogdro:
Thank you.
Just like with https://github.com/veracrypt/VeraCrypt/pull/1025 - I didn't have the time to investigate this more. But, if the problem is fixed, whatever the way, I'm glad.

idrassi:
I have started experimenting with ChatGPT (GPT-4) and so I decided to see if it can find a fix for this issue. Since it cannot read the whole source code of VeraCrypt, I helped it by giving it the source code of the method TextUserInterface::MountVolume and telling it that the issue is in this method. After 2 seconds, it was able to implement the correct fix by setting the option.PIM to -1 in the correct places.

This is not very impressive since this is a simple case but it shows that maybe I should use ChatGPT (especially GPT-4) as a kind of reviewer to get a second option on specific changes.

Below are screenshots of the discussion:

image

image

image

image

bogdro:
Interesting :)
I still prefer to write my own code, though, so that I know I'm the author :). And for the fun of learning.

PR #1029, opened 2023-02-01
Fix verbose message incorrectly saying "dismounting" when mounting on Linux

Jertzukka:
Currently on Linux when mounting a volume with -v verbose flag on, regardless of mounting or -d dismounting, you'll get the message of `Volume {0} has been dismounted.`. If I understood this correctly, in `UserInterface.cpp` -> `UserInterface::ProcessCommandLine` incorrectly uses `LINUX_VOL_DISMOUNTED` LangString when mounting with verbosity on.

I couldn't find a fitting existing LangString that conveyed the "has been mounted", so I created a new one called `LINUX_VOL_MOUNTED`.


idrassi (approved):

idrassi:
Thank you for the fix.

Issue #814, opened 2021-08-11
Several minor issues: typo, desktop file updates, etc


unit193:
Howdy,

First off, thanks for maintaining VeraCrypt, as always.

**Issue <span>#</span>1: typo in source**

While looking at the new language support, I noticed there's a typo on line 1021 of src/Main/GraphicUserInterface.cpp
`Gui->ShowInfo (LangStrin["LINUX_VC_RUNNING_ALREADY"]);`
should instead be
`Gui->ShowInfo (LangString["LINUX_VC_RUNNING_ALREADY"]);`

While I'm on it, a lot of the IDM_ vs HK_ strings are just about the same, though I notice the DE translation of IDM_MOUNT_FAVORITE_VOLUMES has &amp; in it.  The latter, HK_, should be for hotkeys or thereabouts, but basically I'm just trying to ensure I add the language support correctly to my indicator patch. :)

**Issue <span>#</span>2: Locale name output**

Another minor one, I noticed when you run VeraCrypt from the console, it now will output the value of $LANG.  This seems like something leftover from debugging.

**Issue <span>#</span>3: Desktop file updates**

So my .desktop file looks a bit different than the one upstream, and I'd like to reconcile that difference.  Mine is below:
```ini
[Desktop Entry]
Type=Application
Name=VeraCrypt
GenericName=VeraCrypt volume manager
Comment=Create and mount VeraCrypt encrypted volumes
Icon=veracrypt
Exec=veracrypt %f
Categories=Security;Utility;Filesystem
Keywords=encryption,filesystem
Terminal=false
MimeType=application/x-veracrypt-volume;application/x-truecrypt-volume;
```
There are several differences: 

1. The exec line is such that it works with mime types.
2. Categories were updated following the spec on [main](https://specifications.freedesktop.org/menu-spec/latest/apa.html#main-category-registry) and [additional](https://specifications.freedesktop.org/menu-spec/latest/apas02.html) categories.
3. As was recommended by the linter, I added a couple keywords.
4. I added MimeType since I also ship /usr/share/mime/packages/veracrypt.xml to associate the file types on Linux.

If you were curious, the contents of said file is:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<mime-info xmlns='http://www.freedesktop.org/standards/shared-mime-info'>
	<mime-type type="application/x-veracrypt-volume">
		<comment>VeraCrypt Volume</comment>
		<glob pattern="*.hc"/>
	</mime-type>
[...]

PR #1064, opened 2023-05-17
Fix mount.veracrypt external helper

Jertzukka:
`mount.veracrypt` is executed with the parameters from `/etc/fstab` when the fs-type is set to veracrypt. This allows one to mount a volume on boot without other tools such as cryptsetup.

This PR reverses order of --non-interactive and --text to actually make it run, also parse --pim correctly.

For testing, example usage in `/etc/fstab`:
`/home/user/volume.hc /media/veracrypt1 veracrypt password=mysecret,pim=0,nofail 0 0`

Though another way to pass in passwords and pim is preferred as these can be seen in process list (see ps(1)). Though when volume is mounted this way on boot, it is not recognized by Veracrypt which seems like a bug.


idrassi (approved):

idrassi:
Thank you for the fix and for adding PIM support.
Indeed, `--text` must always come first, I don't know how we overlooked this during the merge of https://github.com/veracrypt/VeraCrypt/issues/814 that was authored by @Unit193!

PR #1071, opened 2023-05-23
Update Libzip to latest 1.9.2

DLL125:
Updated the Libzip library to latest version 1.9.2 and changed the version number in the config.h from 1.7.3 to 1.9.2. Not sure if anything else needs to be tweaked :)

### DIFF
--- a/src/Common/libzip/LICENSE
+++ b/src/Common/libzip/LICENSE
@@ -1,6 +1,6 @@
 Copyright (C) 1999-2020 Dieter Baron and Thomas Klausner
 
-The authors can be contacted at <libzip@nih.at>
+The authors can be contacted at <info@libzip.org>
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions
--- a/src/Common/libzip/NEWS.md
+++ b/src/Common/libzip/NEWS.md
@@ -1,3 +1,35 @@
+1.9.2 [2022-06-28]
+
+* Fix version number in header file.
+
+1.9.1 [2022-06-28]
+===================
+
+* Fix `zip_file_is_seekable()`.
+
+1.9.0 [2022-06-13]
+==================
+
+* Add `zip_file_is_seekable()`.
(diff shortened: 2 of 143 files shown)


idrassi:
Unfortunately there are compilations errors (both VS2019 and VS2010) with this PR changes caused by the use of `_Nullable` keyboard in libzip.
image

This is actually known integration issue affecting new libzip version on Windows and libzip headers must be tweeked to allow compilation.

There is a Github project that seems to provide modified libzip that is compatible with Visual Studio 2019 and 2010: https://github.com/kiyolee/libzip-win-build

Is it possible for you to adapt your PR based on what is done on this project?

DLL125:
Ok, I've made the changes, hope it works! :)

DLL125:
I also want to share this fork of zlib with you, https://github.com/zlib-ng/zlib-ng
Seems to be a more moster and more optimized variant.

idrassi:
There are still many compilation errors caused by missing zipconf.h file.
To me it looks like you didn't try to build VeraCrypt with your changes.
Actually, zipconf.h is present on the fork I shared alongside config.h: they are both located under "win32" folder at https://github.com/kiyolee/libzip-win-build/tree/main/win32

I appreciate your efforts but you should first verify your changes locally in order to ensure the quality of the PR. I think it is almost ready, you just need to add missing files and check that it builds cleanly.

Concerning, zlib-ng project, it is indeed an optimized fork but since VeraCrypt usage of zlib is limited and its use case doesn't require high performance, I prefer to stick with the standard zlib since it is more mature and well-established in the open source word. This would have been a different story if VeraCrypt used zlib for performance demanding tasks.

DLL125:
My bad I missed those.... You know when not sleeping to much :). It should work now...

idrassi:
There are still build errors because the file zip_err_str.c was not updated and some new files were not added to the VS projects which caused link errors.
But it is ok, I will merge this PR and do the changes myself.

PR #1076, opened 2023-05-25
Update Polish translation

Matthaiks:
Newest string + some fixes.


idrassi:
Thank you for the update

PR #1078, opened 2023-05-25
Hide failing fs options in --text --create

Jertzukka:
Removes the options exFAT and Btrfs being shown when creating a volume in text mode when the system does not support them and will end up erroring out at the end. Hide Btrfs option when the volume is too small, as we will anyways fail right after.

Hardcoded numbering changed to dynamic as the available options are not necessarily consecutive.


idrassi (approved):

idrassi:

PR #1081, opened 2023-05-27
libgcrypt_test_1

wendig0x:
Test implementation of some cryptographic functions on libgcrypt library

### DIFF
--- a/src/Common/Dlgcode.c
+++ b/src/Common/Dlgcode.c
@@ -13,6 +13,8 @@
 
 #include "Tcdefs.h"
 
+#include "gcrypt.h"
+
 #include <windowsx.h>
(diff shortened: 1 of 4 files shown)


idrassi:
Thank you for this PoC.
The main issue with libgcrypt is that we cannot use for Windows driver: the Windows driver doesn't allow dynamic linking to non-system libraries and so static linking third-party libraries is mandatory but we cannot statically link libgcrypt because of incompatibilities between Apache 2.0 license, TrueCrypt license and LGPL license.

The best alternative to libgcrypt is Botan library (https://github.com/randombit/botan) since it has a more permissive license. Integrating Botan was one of the proposal of the BSI security audit report.

PR #1082, opened 2023-05-27
fix default algos

wendig0x:
Removed the default values for the algorithms. The default values attack first.

https://blog.elcomsoft.com/2021/06/breaking-veracrypt-obtaining-and-extracting-on-the-fly-encryption-keys/

### DIFF
--- a/src/Format/Tcformat.c
+++ b/src/Format/Tcformat.c
@@ -4153,6 +4153,8 @@ BOOL CALLBACK PageDialogProc (HWND hwndDlg, UINT uMsg, WPARAM wParam, LPARAM lPa
 				int ea, hid;
 				wchar_t buf[100];
 
+				srand(time(NULL));
+
 				// Encryption algorithms
 
 				SendMessage (GetDlgItem (hwndDlg, IDC_COMBO_BOX), CB_RESETCONTENT, 0, 0);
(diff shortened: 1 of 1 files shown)


idrassi:
Thank you.
I understand the idea behind this change but the choice of the algorithm can have impact on usability depending on the usage context (for example AES/SHA512 is the fastest combinaition and this is what most users look for). I see this feature as more towards security-focus people. Probably we should add a setting to enable randomization of algorithms during encryption process. 

Concering the implementation, I would not use CRT srand/rand. I prefer to use more secure alternative (like [RtlGenRandom](https://learn.microsoft.com/en-us/windows/win32/api/ntsecapi/nf-ntsecapi-rtlgenrandom))

wendig0x:
This change leaves the choice of algorithm to the user. Many people just don't make any choice, and make it easy for attackers to attack. As for RtlGenRandom - I think in this case, srand does not affect security.

wendig0x:
@idrassi 

I found another confirmation that attackers are using default values to crack VeraCrypt:
https://www.forensicfocus.com/articles/how-to-efficiently-decrypt-truecrypt-veracrypt-encryption-using-passware/

I think I've found a compromise solution: maybe just add a text warning that the default values used help the attacker?

PR #1085, opened 2023-05-29
Update Corsican translation on 2023-05

Patriccollu:
Hello,

 * https://github.com/veracrypt/VeraCrypt/commit/8072207eb6ad0783a4e2c54addc4b3196738029e Add a new LangString 'LINUX_VOL_MOUNTED' to show when volume mounted.

I also updated several existing strings to conform to Corsican punctuation rules. This includes replacing a normal space by a non-breaking space (Alt+'0160') in front of some punctuation signs - colon, semicolon, question mark, exclamation mark, and right double angle quotes, and also behind left double angle quotes - in order to avoid having such a sign alone on a line after a break triggered by unavailable room.

Best regards,
Patriccollu.


idrassi:

PR #1086, opened 2023-05-29
Fix issues launching fsck via terminal on Linux

Jertzukka:
Currently on a system without xterm or konsole (like fresh Ubuntu install) the fsck check (Check filesystem, repair filesystem) will not launch. Fixed an issue where konsole will always error out as `--caption` is not a valid argument (anymore?) and `--title` is a no-op for compatibility. Now we use `-p tabtitle=` profile property override instead. Also added gnome-terminal as an alternative, though I couldn't figure how to get gnome-terminal launching without requiring dbus-x11 so we make sure the system has it (tried suggestions such as --disable-factory but no help). If anyone knows a way to not depend on it, feel free to pitch in.

Previously the error message was simply "xterm not found" when neither xterm nor konsole were found, so a new Exception `TerminalNotFound` and a new LangString `LINUX_EX2MSG_TERMINALNOTFOUND` was added to let the user know which program/package they need.


idrassi:
Thank you, this is a good enhancement.
Unfortunately there is a conflict with the file "Translations/Language.co.xml" because translation PR was merged and that changes tabs/spaces on the file.
Can you please help resolve this conflict?

PR #1087, opened 2023-05-30
Format Corsican localization XML file

Patriccollu:
Hello,

This is an update of **Corsican** (co) localization to standardize the format of the whole XML file by using the '_Pretty print - indent only_' function of Notepad++'s plugin XML Tools. In a nutshell, it makes the XML code easier to read and reduces the file size by using tabs instead of spaces as indent.

For your information, this function was recently used to [standardize the format of all Notepad++ localization files](https://github.com/notepad-plus-plus/notepad-plus-plus/pull/13631).

Best regards,
Patriccollu.


idrassi:
Thank you.
This is an interesting approach

PR #1089, opened 2023-06-01
Update Polish translation

Matthaiks:
New string.


idrassi (approved):

idrassi:
Thank you for the update

PR #1100, opened 2023-06-07
Linux/FreeBSD: Fix privilege escalation prompts being ignored

Jertzukka:
Currently if you fail the privilege escalation prompt, the second one and consecutively every second will be ignored. This is because if we do not use --use-dummy-sudo-password and are on Linux/FreeBSD, we will be prompted for password twice for each evaluation in the while(!ElevatedServiceAvailable) loop.

For the fix, we make sure that we run the prompt only once for each case.


idrassi:
Thank you for the fix of this 4 years old bug!
Indeed, the negative case was not handled properly.

PR #1103, opened 2023-06-11
Fix WXSTATIC building on latest wxWidgets stable 3.2.2.1

Jertzukka:
I was writing very elaborate Makefile parsing the wxWidgets version number from $(WX_ROOT)/include/wx/version.h to add/remove ./configure flags based on the version number, until I realized it would become quite crowded if we keep supporting the working VC_WX_MINIMAL flags for older wxWidget versions. I already have it mostly working, but I'm not sure if there is point to it. If we just care about the minimal build working on the latest stable, this MR fixes it.

Changes to be able to wxbuild with latest wxWidgets stable 3.2.2.1.

Removed in 3.1.5, needs to be removed.
--disable-webkit

Added in 3.1.0, needs to be added.
--disable-addremovectrl

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -453,7 +453,7 @@ WX_CONFIGURE_FLAGS += --enable-unicode -disable-shared --disable-dependency-trac
 
 ifdef VC_WX_MINIMAL
 WX_CONFIGURE_FLAGS += --disable-protocol --disable-protocols --disable-url --disable-ipc --disable-sockets --disable-fs_inet --disable-ole --disable-docview --disable-clipboard \
-	--disable-help --disable-html --disable-mshtmlhelp --disable-htmlhelp --disable-mdi --disable-metafile --disable-webkit --disable-webview \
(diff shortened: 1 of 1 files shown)


idrassi:
Thank you for this change.
I still need to do some tests because currently official builds still use wxWidgets 3.0.5 and the generic installer is built under CentOS 6 for maximum compatibility.
I will check how things go with wxWidgets 3.2.2.1.

idrassi:
I have found two issues when building static wxWidgets 3.2.2.1 library :
- if the machine has libcurl devlopement libraries, the target wxbuild fails because we have disabled sockets support
- wxWidgets 3.2.2.1 defaults to GTK3 but VeraCrypt makefile expects the default to be GTK2 which causes a failure of wxbuild target

I have a fix for these issues (disabling curl support for wxWidgets and explicitely set GTK2 as default if WITHGTK3 is not specified). So, I will merge this PR and then I will commit the fixes.

PR #1120, opened 2023-06-23
Dll125 lzma

DLL125:
Updated LZMA to latest version.

Note: "7zWindows.h" most likely needs to be added to the VS project files.


idrassi:
Thank you for the update.
Adding "7zWindows.h" to the project files is not needed for the build to succeed but it is better for easy access to it. I will take it care of it after merging this PR.

PR #1136, opened 2023-06-29
Linux: Add libpcsclite-dev dependency into compilation guide

Jertzukka:
After the EMV support commits, the project won't build anymore without the new dependency of libpcsclite-dev.

Probably also needs to be added into CMakeLists.txt as dependency, but I'm not too familiar with that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ Requirements for Building VeraCrypt for Linux and Mac OS X:
   wxWidgets 3.0 library source code (available at https://www.wxwidgets.org)
 - FUSE library and header files (available at https://github.com/libfuse/libfuse
   and https://osxfuse.github.io/)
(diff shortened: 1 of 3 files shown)


idrassi:
Thank you for your insightful contribution!

Concerning the modification of CMakeLists.txt, it's important to clarify that VeraCrypt doesn't explicitly link against pcsclite. It loads the pcsclite library dynamically at runtime when necessary. This dynamic loading is the reason we require the pcsclite-dev package, specifically for its header files.

However, I see the practical benefit in specifying pcsclite as a dependency. By doing this, the EMV feature will become available by default upon installation. Without this step, users would need to perform the manual installation process. Thus, I will add pcsclite as a dependency in CMakeLists.txt.

PR #1144, opened 2023-07-02
Renormalize 7zWindows.h from CRLF to LF line separator as intended

Jertzukka:
This file is supposed to have LF line separator instead of CRLF as per .gitattributes, this causes the file to always show as modified in Git.


idrassi:
Indeed!
Thank you for the fix.

Jertzukka:
1 minute turnaround time is fast. Thanks :+1:

PR #1147, opened 2023-07-04
Linux: Overwriting file container allows you to use its previous space

Jertzukka:
When replacing a file container, we increase the `diskSpace` by the file container's size. This doesn't affect the hidden volume space logic, as in that case we use `MaxVolumeSize` when `MaxVolumeSizeValid` is true instead of `AvailableDiskSpace`.


idrassi:
Thank you for this fix. I should implement similar logic for Windows

PR #1154, opened 2023-07-20
Add note about Linux usage to docs

springfielddatarecovery:
Docs specify cascading encryption should be specified as "AES(Twofish)" but on Linux it requires "AES-Twofish". This also is not documented by the --help flag but I was unable to find where to update that.

### DIFF
--- a/doc/html/Command Line Usage.html
+++ b/doc/html/Command Line Usage.html
@@ -239,7 +239,7 @@ It must be followed by a parameter indicating the PRF hash algorithm to use when
 <tr>
 <td>/encryption</td>
 <td>(Only with /create)<br>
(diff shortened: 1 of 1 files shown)


Jertzukka:
Yeah, they're only mentioned in the Cascades.html but never specified the format to be used for Linux command-line usage which probably should be documented better. That part of the docs is specifically for Windows and mentioning it under VeraCrypt Format.exe doesn't seem helpful. The best place to put it would be the `--text --help` output here:

https://github.com/veracrypt/VeraCrypt/blob/c0785aa1d4b6d988f175ca5732641e6ae2b10cc7/src/Main/UserInterface.cpp#L1213-L1215

Maybe something along the lines of "When cascading ciphers, the ciphers are separated with a dash e.g. AES-Twofish-Serpent."

springfielddatarecovery:
Good point, I'll close this one and I have submitted a separate one for that file.

PR #1155, opened 2023-07-21
Update UserInterface.cpp

springfielddatarecovery:
Docs specify cascading encryption should be specified as "AES(Twofish)" but on Linux it requires "AES-Twofish".

### DIFF
--- a/src/Main/UserInterface.cpp
+++ b/src/Main/UserInterface.cpp
@@ -1211,7 +1211,7 @@ namespace VeraCrypt
 					" Display password characters while typing.\n"
 					"\n"
 					"--encryption=ENCRYPTION_ALGORITHM\n"
-					" Use specified encryption algorithm when creating a new volume.\n"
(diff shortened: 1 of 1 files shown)


idrassi:
Thank you for the contribution. Your change in the help message aligns with the current implementation in the Linux environment and certainly adds clarity for our Linux users.

In the Windows environment, I have opted for the 'AES(Twofish)' notation to mirror the concept of function composition in mathematics. The intent was to visually represent that the plaintext is first encrypted using Twofish, and then that output is encrypted once more with AES.

On the other hand, the 'AES-Twofish' notation could be ambiguous. Without additional context or explanation, a user may understandably infer that AES is the initial encryption method followed by Twofish.

Moving forward, it's critical to have uniformity across platforms in how we represent cascading algorithms. To this end, I will improve the Linux version to support the parenthetical notation while keeping support for the current notation.

PR #1160, opened 2023-07-29
Windows: Add installer language support for rest of the translations

Jertzukka:
When I was working on one of the translations and wanted to test it out in the correct context, I noticed that language never is offered as an option in the Setup installer. Currently there are a lot of translated strings that are never seen anywhere as the installer only offers select few languages.

This PR adds in rest of the languages available in Translations/, apart from Burmese which I couldn't find a correct LANG_ definition for from WinNT.h.


idrassi:
Interesting contrubution, thank you!
At the time, I focused on languages that had most translated resources because some languages had minimal tranlations and their were no active translator for them.
Now there are more languages with enough level of translation.
I will merge this PR and we will see in the future how to manage languages that lack enough translated strings.

PR #1164, opened 2023-08-03
fix warnings and UB

kovalev0:


idrassi:

PR #1167, opened 2023-08-07
Call EnsureVisible() for OSX in OnActivate

denizt:
When using multiple monitors, the resolution can change in OSX, during plug-in/off off the display(s). To avoid window becoming non-usable, there is EnsureVisible, that is called in several places.

However, if you minimize VeraCrypt on the bigger screen, and restore it in the main screen of a MacBook, after unplugging, the window will become unusable (unless you know that clicking 'About' calls EnsureVisible :-)).

Call EnsureVisible OnActivate so MainFrame is always functional across screens, even when minimized.


idrassi:
Thank you for you analysis and for the fix. I really appreciate!

PR #1168, opened 2023-08-08
Update Language.ro.xml

TigerxWood:


Translations/Language.ro.xml
```diff
@@ -1388,251 +1388,251 @@
...
+    <entry lang="ro" key="EXPANDER_WRITING_RANDOM_DATA">Se scriu datele aleatorii în spațiul nou...\n</entry>
+    <entry lang="ro" key="EXPANDER_WRITING_ENCRYPTED_BACKUP">Se scrie copia de rezervă a antetului recriptat...\n</entry>
+    <entry lang="ro" key="EXPANDER_WRITING_ENCRYPTED_PRIMARY">Se scrie antetul primar recriptat...\n</entry>
+    <entry lang="ro" key="EXPANDER_WIPING_OLD_HEADER">Se distruge vechiul antet de rezervă...\n</entry>
+    <entry lang="ro" key="EXPANDER_MOUNTING_VOLUME">Se montează volumul...\n</entry>
+    <entry lang="ro" key="EXPANDER_UNMOUNTING_VOLUME">Se demontează volumul...\n</entry>
+    <entry lang="ro" key="EXPANDER_EXTENDING_FILESYSTEM">Se extinde sistemul de fișiere...\n</entry>
   </localization>
-  <xs:schema attributeFormDefault="unqualified" elementFormDefault="qualified" xmlns:xs="http://www.w3.org/2001/XMLSchema">
+  <xs:schema attributeFormDefault="unqualified" elementFormDefault="qualified" xmlns:xs="http: /  / www.w3.org / 2001 / XMLSchema">
```
Jertzukka, line 1635: It seems you have done some kind of find and replace on '/' to ' / ', I don't know if that is the grammar norm in Romanian, but it seems to have caught an URL here.

idrassi: I agree with @Jertzukka. There so many places where spaces were added before and after `/` character. The URL part is definitely wrong but other places also seem suspecious. @TigerxWood can you please clarify why so many spaces were added around `/` character occurences?

TigerxWood:
Yes. indeed the new form of the translation is the correct one with spaces before and after '/' and even if I checked again all strings from the file it seems I have stoped where the translation ends. And also because when I checked the translation with the beta 1.26.4 I did not received any error it  seems I missed it. So, I will update again with correction made.

idrassi:
Thank you @TigerxWood for the clarification and for the correction.

Issue #1171, opened 2023-08-11
Veracrypt -c is unusable on Alpine due to weird text printing issue


Sod-Almighty:
**VeraCrypt version:** 1.26.4
**Operating system and version:** Alpine Linux 3.18
**System type:** 64-bit

I compiled Veracrypt on Alpine linux as detailed [here](https://github.com/veracrypt/VeraCrypt/issues/1165#issuecomment-1665585789), and it just isn't usable. It's having severe difficulties in printing messages to the console.

For example, I tried to create a volume:

```
# veracrypt -c test
Illegal option -p
Volume type:
 1) Normal
 2) Hidden
```

(It prints `Illegal option -p` because I haven't recompiled since [this commit](https://github.com/veracrypt/VeraCrypt/commit/dab261c6941e1ee04d823effdf65784f1a554cb8).)

So, I type "1" and enter. The terminal now looks like this:

```
# veracrypt -c test
Illegal option -p
Volume type:
 1) Normal
 2) Hidden
1
Select [1]:

```

Weird right? So I type "1" again, and press enter. The terminal now looks like this:

```
# veracrypt -c test
Illegal option -p
Volume type:
 1) Normal
 2) Hidden
1
Select [1]:
1
Enter volume size (sizeK/size[M]/sizeG.sizeT/max):
Encryption Algorithm:
 1) AES
 2) Serpent
 3) Twofish
 4) Camellia
 5) Kuznyechik
 6) AES(Twofish)
 7) AES(Twofish(Serpent))
 8) Camellia(Kuznyechik)
 9) Camellia(Serpent)
 10) Kuznyechik(AES)
 11) Kuznyechik(Serpent(Camellia))
 12) Kuznyechik(Twofish)
 13) Serpent(AES)
 14) Serpent(Twofish(AES))
 15) Twofish(Serpent)

```

Veracrypt has taken my input as being the desired size of the file! Before it even asked the question!

Something is seriously awry.

Jertzukka:
There seems to be some difference between libc and musl on how/when they flush the stdout buffer. You can workaround it by changing:

https://github.com/veracrypt/VeraCrypt/blob/f48c21a54ba461c0356d5a9358ae7e441ab269c3/src/Main/TextUserInterface.cpp#L1048-L1051

[...]

PR #1172, opened 2023-08-12
Linux: Flush stdout explicitly when reading stdin

Jertzukka:
Rules of automatic flushing of stdout buffer is implementation-defined behaviour. In glibc this is automatically flushed, but we can't rely on it for other implementations such as musl. More info: https://stackoverflow.com/questions/39536212/what-are-the-rules-of-automatic-stdout-buffer-flushing-in-c/39536803#39536803

Fixes: https://github.com/veracrypt/VeraCrypt/issues/1171


idrassi:
Wow, amazing! I was not aware of this flushing behavior.
Thank you for this enhancement.

PR #1175, opened 2023-08-14
Update Language.zh-cn.xml

nkh0472:
- translate for f84d235cf17a92bb51031833da502660d364013f
- unify the translation of "readonly"


idrassi:
Thanks

PR #1177, opened 2023-08-16
Update Language.zh-cn.xml

cjl3230:
https://github.com/veracrypt/VeraCrypt/blob/d3ecffac3b46abb0d6a91a82838c05b11fd6fe0c/Translations/Language.zh-cn.xml#L1608C89-L1608C89 Missing character n after backslash


idrassi:
Thanks for the fix!

PR #1181, opened 2023-08-19
Update zlib to latest

DLL125:
Updated zlib to latest version 1.3.


idrassi:
Thank you for the update

PR #1182, opened 2023-08-21
Allow cross compilation.

Unit193:
This fixes up pkg-config such that cross compilation is successful.

Note: I only tested this on Linux, specifically Debian.

### DIFF
--- a/src/Driver/Fuse/Driver.make
+++ b/src/Driver/Fuse/Driver.make
@@ -15,6 +15,6 @@ NAME := Driver
 OBJS :=
 OBJS += FuseService.o
 
-CXXFLAGS += $(shell pkg-config fuse --cflags)
+CXXFLAGS += $(shell $(PKG_CONFIG) fuse --cflags)
 
 include $(BUILD_INC)/Makefile.inc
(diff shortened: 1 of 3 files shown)


idrassi:
Thank you for this enhancement. This looks generic enough to me.

Unit193:
It seems for the 1.26.7 release this regressed due to the new PCSC support, specifically on the following lines:
https://github.com/veracrypt/VeraCrypt/blob/34407f4b31d59fe61d2bed639e7ff542db29a20c/src/Makefile#L196
https://github.com/veracrypt/VeraCrypt/blob/34407f4b31d59fe61d2bed639e7ff542db29a20c/src/Makefile#L382
https://github.com/veracrypt/VeraCrypt/blob/34407f4b31d59fe61d2bed639e7ff542db29a20c/src/Makefile#L443
https://github.com/veracrypt/VeraCrypt/blob/34407f4b31d59fe61d2bed639e7ff542db29a20c/src/Makefile#L470

idrassi:
Indeed! However, these lines were already present when you submitted the PR. I hadn't considered checking the pkg-config usage beyond the Ayatana library... we both overlooked it!
I'll modify these lines based on your initial PR.

Unit193:
Haha, oh dear.  That's what I get for making the change on the release tarball then porting the changes over to git.  Yeah that's entirely my fault, thanks for looking into it.

PR #1190, opened 2023-09-03
Update Language.fi.xml

Jertzukka:
The current Finnish translation is in a very poor state and this PR fixes most of it, but not all. I'll continue with it at some point in the future, but will submit these changes meanwhile as it is definitely a huge improvement.

Line that has not been edited doesn't mean I necessarily approve of the current translation, it is also possible that I haven't just had the time to look at it.


idrassi:
Thank for the time and effort you put to enhance the Finnish translation. I really appreciate!

PR #1192, opened 2023-09-07
Update Language.xml

TigerxWood:
Corected a typo


idrassi:
Thank you for the correction.

PR #1196, opened 2023-09-08
Update Corsican translation on 2023-09

Patriccollu:
Hello,

 * https://github.com/veracrypt/VeraCrypt/commit/f15052e68d43bdad8d8a125a3601ff8cedcf5a3d Windows: Add link in keyfiles dialog to documentation page for risks of third-party file extensions usage.

I also moved `</localization>` statement at the correct place because this was no longer the case since June 29th!

Cheers,
Patriccollu.


idrassi:
Thank you for the update and the fix. I diidn't spot the </localization> statement before this!

PR #1200, opened 2023-09-13
Update Language.it.xml

robo3945:
Some italian translation update. I'm a native italian.


idrassi:
Thank you for the updated Italian translation. I don't speak Italian but I checked with some online tools and they confirm that your translation is better.

PR #1203, opened 2023-09-20
Update Corsican translation on 2023-09 (2nd)

Patriccollu:
Hello,

 * https://github.com/veracrypt/VeraCrypt/commit/0f3ae268a4b0dfac7090f4fbf969601a1c35c86f Windows: Add setting in main UI and setup wizard to disable memory protection

Cheers,
Patriccollu.


idrassi:
Thank you for the swift update

PR #1183, opened 2023-08-21
Update Language.zh-cn.xml

fzxx:
1. 把所有”装载“替换为”加载“，统一词语
2. 添加易于理解三种不同创建方式的提示

PR #1206, opened 2023-09-22
Update Language.zh-cn.xml

fzxx:
1、翻译
2、统一加载按钮的翻译


Translations/Language.zh-cn.xml
```diff
@@ -1573,7 +1573,7 @@
     <entry lang="zh-cn" key="IDC_USE_ALL_FREE_SPACE">使用所有空闲空间。</entry>
     <entry lang="zh-cn" key="SYS_ENCRYPTION_UPGRADE_UNSUPPORTED_ALGORITHM">VeraCrypt无法升级，因为系统分区/驱动器使用了将不再支持的算法进行加密。\n请在升级VeraCrypt之前解密系统，然后再次加密。</entry>
     <entry lang="zh-cn" key="LINUX_EX2MSG_TERMINALNOTFOUND">未找到支持的终端应用程序，您需要xterm、konsole或gnome终端（带有dbus-x11）。</entry>
-    <entry lang="zh-cn" key="IDM_MOUNT_NO_CACHE">无缓存装载</entry>
+    <entry lang="zh-cn" key="IDM_MOUNT_NO_CACHE">无缓存加载</entry>
```
nkh0472, line 1576: As we discussed earlier in #1183, keeping the translation for "mount" as "装载" is better...

fzxx: explorer_TTdXZABe1A
explorer_I3OW4USVJJ
全部更改或许不合适，但加载按钮需要统一

fzxx: @idrassi 你认为呢？

idrassi: Indeed, there is a descrepency, thank you for pointing this out.
I agree: we should have the same translation for "Mount" and "Mount without cache" in this button.
I noticed that there are many UI menu entries that also use "加载". So the quickest thing to do is to use "加载" for IDM_MOUNT_NO_CACHE since I want to buid official release today.

@fzxx : I cannot merge your PR because of a conflict. I will try to solve it.

idrassi:
I will now able to merge this PR. Thanks.

PR #1216, opened 2023-10-01
Update Language.uk.xml

udev2045:


idrassi:
Thank you for the contibution.
Can you please share expalanation about the changes you made so that others can have better understanding during review?
Thanks.

udev2045:
Corrected words with the letter **ы** which is not in the Ukrainian alphabet.

idrassi:
@udev2045 Thank you for the exlanation but I have a doubt concerning `IDC_PAUSE` translation: based on various online translation tools "Павза" seems to be a better translation for "Pause" and so the original translation looks ok (actually I could not find a refrence for "Навза" online).

What do you think?

Other than this, all other translations seem ok and I can merge them. So only `IDC_PAUSE` is on hold.

PR #1222, opened 2023-10-02
Fix Finnish language pack parsing issue

Jertzukka:
A missing 'n' for a new line character causes the parsing of the language pack to fail. I remember noticing this and fixing it at some point, but it must have gotten lost in git reverts while I worked on this over the months. I didn't unfortunately notice the issue before the final PR submission as I was working on it on Linux and this issue only shows up on Windows.


idrassi:
Thank you for the fix. It is an unfortunate issue and it is not the first time that it happens with language files.
I should systematically try to load all languages to detect such problems early but everytime I lack time to do such tests.
Managing properly a software with the magnitude of VeraCrypt is not simple.

PR #1226, opened 2023-10-03
fix for corrupted icons on macOS

sardanap:
This should fix corrupted icons on macOS without (hopefully) breaking anything else 😄 
Be patient this is my 1st PR at all


idrassi:

PR #1229, opened 2023-10-08
Fix warnings and throwing an exception instead of ignoring the error

kovalev0:


idrassi:

PR #1230, opened 2023-10-08
Update Language.uk.xml

udev2045:
Spelling corrections have been made and English translations have been added. Checked lines 1 through 400.

### DIFF
--- a/Translations/Language.uk.xml
+++ b/Translations/Language.uk.xml
@@ -11,10 +11,10 @@
     <entry lang="uk" key="IDC_BROWSE">О&amp;гляд...</entry>
     <entry lang="uk" key="IDC_DESKTOP_ICON">&amp;Значок VeraCrypt на Рабочому столі</entry>
     <entry lang="uk" key="IDC_DONATE">допомога проекту...</entry>
-    <entry lang="uk" key="IDC_FILE_TYPE">Ассоціювати роз&amp;ширення '.hc' с VeraCrypt</entry>
+    <entry lang="uk" key="IDC_FILE_TYPE">Ассоціювати роз&amp;ширення '.hc' з VeraCrypt</entry>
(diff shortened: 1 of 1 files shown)


idrassi:
Like the other PR, I did a quick check using online tools and it appears there are two entries that may have problems:

- IDC_PAUSE: original "Павза" seems to be more standard than proposed "Паyза" (internet search of "Паyза" doesn't give any meaningful results while "Павза" is used overwhelmingly)
- IDM_RESUME_INTERRUPTED_PROC: both "прерваний" and "перерваний" seem correct but it looks like original "прерваний" is more modern and so more adequate.

What do you think?

udev2045:
Hi Friend
Thank you for your comment.

- IDC_PAUSE:

Both 'Павза' and 'Паyза' are correct, but the word 'Паyза' has been used all my life.
The word 'Павза' was introduced into the Ukrainian language on June 3, 2019 as a variant spelling of the word 'Паyза'
Read more at https://vnu.edu.ua/uk/articles/ukrayinskiy-pravopis-2019-osnovni-zmini-ta-dopovnennya
You can check the correct spelling of words here https://slovnyk.ua/
So I think it is better to use the word 'Паyза' which is used everyday and everywhere than to use a new word which is still very rarely used. 


- IDM_RESUME_INTERRUPTED_PROC:

The word "прерваний" does not exist in the Ukrainian language.
The correct word is "перерваний."
Perhaps someone made a translation from Russian and got the wrong translation
It's Russian for "прерваный".  https://i.imgur.com/2SzJFdp.png     https://i.imgur.com/s4UMjy0.png
In Ukrainian, "перерваний".  https://i.imgur.com/32EV9bC.png

Best regards, Ivan

idrassi:
Thank you @udev2045 for all these clarification.
I will merge this PR but what about the previous one? will you close it?

PR #1239, opened 2023-10-30
Linux: Focus PIM field when selected

### LINKED ISSUE #1238: Wrong keybord focus control in Linux GUI when mount volume with PIM
If you activate the "Volume-PIM" checkbox via the spacebar in the mount dialog, the keyboard focus jumps to a different location. In the Windows version, the focus is set into the PIM field. 

Please change the focus behavior in the Linux GUI to the behavior in Windows. 

Linux Debian 12.2 (64 bit)/ Ubuntu 22.04 LTS (64 bit)
VC 1.26.7

Thanks a lot

Jertzukka:
Improve accessibility by focusing VolumePimTextCtrl initially when the checkbox is selected.

Fixes #1238


idrassi:
Thank you for the enhancement.

PR #1241, opened 2023-11-02
02Nov23-Update Language.es.xml

MayanTigger:
I'm not sure of the changes will remain, two times of I have to re do the translations. Please confirm of the translations are "as it should be".


idrassi:
The translations are in the correct format, so no issue on this side.
I will merge your PR.

MayanTigger:
Great!
Thanks for the feedback, Then I will continue with the translations

PR #1244, opened 2023-11-07
Linux: Fix generic installation script on Konsole in Wayland

Jertzukka:
Currently the generic installation script will break on Konsole in a Wayland session as the x11 aliases such as --title are only available in the Qt application if the XDG_SESSION_TYPE is x11. Instead of using an alias, we can use --qwindowtitle directly.

Not tested on FreeBSD but should apply similarly.


idrassi:
Thank you for this fix. Since "--qwindowtitle" is linked to QT, I agree that it should work on FreeBSD too. I will test it later and if I find an issue, I will report it.

Issue #1201, opened 2023-09-17
The macOS version can only display the English interface, and other languages cannot be set. Please fix it. Thank you.


vitaminx:
As shown below, no language other than English can be selected👇

![image](https://github.com/vitaminx/pic/blob/main/github/VeraCrypt-1.png?raw=true)

But there are language files in the installation package👇

![image](https://github.com/vitaminx/pic/blob/main/github/VeraCrypt-2.png?raw=true)

idrassi:
VeraCrypt supports now selecting language to use for its GUI (configurable through the menu Settings - Language).

PR #1253, opened 2023-11-15
Linux/FreeBSD/macOS: Implement language selection settings

### LINKED ISSUE #875: Change program language for Linux
Hi,
I would like to change the language of the VeraCrypt interface, how can I do that?
I tried to add the string ``<config key="Language">en</config>`` to the ``Configuration.xml`` file but it doesn't work.

Version of VeraCrypt installed: v1.25.4 (downloaded from https://veracrypt.fr/en/Downloads.html)
OS: Ubuntu 20.04

Thanks

Jertzukka:
Currently the only way to have different language in VeraCrypt than on your system is to set env variables to your shell before launching the executable, but this is not very user-friendly. This PR implements a new tab in the PreferencesNotebook for selecting a language pack to use. Existing users won't notice a difference as by default it uses the existing behaviour of using your LANG env variable. A new entry is added into Configuration.xml which defaults to "system" to save your preferred language.

Changes:
- New Language tab in the preferences menu
- Shortcut to Language tab in MainMenuBar's Settings
- Mapping translator and current language pack in `LanguageStrings.cpp` so they can be used in the GUI

Translations:
- 3 new translation entries, `LINUX_LANGUAGE` entry borrowed from `IDM_LANGUAGE` where it looked correct.
- Few translation changes in Finnish & Korean translations
- Change the native language name of zh-hk to differ it from zh-tw in `Setup.c` so we don't have two languages with the same exact name

Adjacent:
- FreeBSD Setup now copies language files to the proper location so they work as expected
- FreeBSD WX minimal build for GTK3 requires wxGraphicsContext, so it is no longer disabled

Sorry for the few unrelated things in this PR, there were some things I stumbled on when doing thing A, so I had to fix B to progress, and didn't really feel like separating them to different PRs.

Currently the new langEntries list in `PreferencesDialog.h` is copied and edited from the list in `Setup.c`. I don't know if there's a good way to share this list as the struct contains Windows types, or if it is just easier to duplicate it.

Let me know what you think about the implementation.

Tested on: Ubuntu 23.04, FreeBSD 13.2, macOS Monterey 12.7.1

Fixes #875 #1201 (Though macOS doesn't detect system language by default as env variables aren't passed to the application when launched through Finder or Dock, requires `launchctl setenv LANG fi_FI.UTF-8` for example.)


idrassi:
Thank you very much for this contribution! I did a first review and things are ok for me. Congratulations for the amazing work :+1:

PR #1260, opened 2023-11-22
src/Setup/Linux: Specify icon in mime info

joebonrichie:
Otherwise file managers will just use a generic missing icon icon.


joebonrichie:
Nautilus for example will not an icon without this patch
Screenshot from 2023-11-22 00-34-36

Jertzukka:
This is what they currently look on my system: 
image

I think this would have to be `<generic-icon name="veracrypt"/>`, or at least on my GNOME system it won't work otherwise. But even so, I think having the application icon being the icon for a container might be confusing. There probably exists some kind of generic icon that represents an encrypted file that could be used if it were to be changed.

Issue #184, opened 2017-07-06
Wayland compatibility


rugk:
It would be nice to make this 100% wayland-compatible.

rugk:
shh #badbot :no_bell::robot::no_bell:

rugk:
shh #badbot :no_bell::robot::no_bell:

DimitriyB:
Are there any news on this issue?

eclairevoyant:
Veracrypt opens fine in wayland. Not sure what is being asked here.

rugk:
> Veracrypt opens fine in wayland. 

[Is it running via the compatibility layer xwayland though?](https://askubuntu.com/questions/1393618/how-can-i-tell-if-an-application-is-using-xwayland)

eclairevoyant:
No, native wayland.

tekstryder:
- Arch Linux | Kernel 6.6.1
- Gnome-shell | Mutter 45.1
- Wayland (`gnome-shell --no-x11`)
- nVidia 545.29.02
- Veracrypt 1.25.9
- wxwidgets 3.2.4

I recently switched my display protocol to a pure Wayland environment (no X11, no Xwayland).

Veracrypt crashes immediately, whether launching the GUI or from CLI.


## Stack Trace:

<details><summary>Coredump</summary>

```
$ coredumpctl gdb 8335
           PID: 8335 (veracrypt)
           UID: 1000 (myusername)
           GID: 1000 (myusername)
        Signal: 6 (ABRT)
     Timestamp: Fri 2023-10-20 17:15:58 EDT (14s ago)
  Command Line: veracrypt
    Executable: /usr/bin/veracrypt
 Control Group: /user.slice/user-1000.slice/user@1000.service/session.slice/vte-spawn-ff8a84be-39c1-4bd0-ba9a-1ce9ed4bd7b4.scope
          Unit: user@1000.service
     User Unit: vte-spawn-ff8a84be-39c1-4bd0-ba9a-1ce9ed4bd7b4.scope
         Slice: user-1000.slice
     Owner UID: 1000 (marc)
       Boot ID: 5b42ada02b1841e3bc8d132f273250da
    Machine ID: b1fa067dbeb84216ab8d4c29c1b78033
      Hostname: myhostname
       Storage: /var/lib/systemd/coredump/core.veracrypt.1000.5b42ada02b1841e3bc8d132f273250da.8335.1697836558000000.zst (present)
  Size on Disk: 847.1K
[...]

PR #1264, opened 2023-11-26
Linux: Allows GUI to launch in a Wayland-only environment

Jertzukka:
Currently we check whether the system has DISPLAY environment variable set, which is the case in a system that uses X11 natively or XWayland. This variable is not set in a system with only Wayland which forces us to text-only mode, so need to also check whether WAYLAND_DISPLAY is set.

Fixes: #184


idrassi:
simple and yet very helpful! Thank you very much 😀

PR #1270, opened 2023-12-03
Update Swedish translation

JonatanWick:
I have reviewed and updated all strings in the Swedish translation. Most strings needed to be updated. Please merge.


idrassi:
Thank you for taking the time to update the Swedish translation.

PR #1273, opened 2023-12-06
macOS: Use correct Disk Utility location when "check filesystem" is ran

Jertzukka:
As of macOS Catalina (version 10.15) some system tools are moved to /System/Applications/. The old path could probably be just replaced with the new location as the older macOS versions are not officially supported, but it is probably safer to just check in case.


idrassi:
Thanks!

Issue #1055, opened 2023-04-26
For Mac, have the option to use FUSE-T


dfloyd888:
# Desired behavior
Please add, as an option for macOS, the ability to use FUSE-T.  FUSE-T is similar to FUSE, but instead of presenting volumes directly to the OS, it presents them as loopback mounted NFS v4 volumes.  Since FUSE-T requires no kernel extensions, it is easier to maintain for MacOS's constantly changing kernel space, and that one has to go into recovery mode to select a lower security option for FUSE, while FUSE-T doesn't require this.

# Additional information

FUSE-T's website:  https://www.fuse-t.org/

# Your Environment

Please tell us more about your environment

**VeraCrypt version:** VeraCrypt 1.25.9

**Operating system and version:** macOS 13.3.1

**System type:** arm64

ZelnickB:
+1 to this issue. FUSE-T markets itself as a "Drop-in replacement for osxfuse," so even though I haven't looked at the VeraCrypt code, I can't imagine that the implementation would be very difficult.

Since VeraCrypt aims to be transparent, I also think that it's problematic that both MacFUSE and FUSE-T are closed-source programs, but I suppose that's a different discussion.

arifzeeshan:
Well,  it will work with minor changes to veracrypt, the developer of Fuse-T did compile and [tested](https://github.com/macos-fuse-t/fuse-t/issues/11#issuecomment-1320898562) it with minor changes.

idrassi:
It is indeed becoming harder to configure macFUSE on recent macOS versions for non-technical users. This limits the userbase of VeraCrypt and supporting Fuse-T is a solution to this problem.

Thank you @arifzeeshan for sharing the confirmation of Fuse-T author that minimal changes are needed to make it compatible. He didn't share these changes but I give it a try in the coming weeks and see how it goes.

arifzeeshan:
Looking forward to the changes and since Fuse-T runs on a userspace, hope we will also have a portable version too.

shadofficial:
Could you please provide an update on the progress made in this area? Any details on the specific changes implemented or the overall experience in integrating Fuse-T with VeraCrypt would be greatly appreciated. Thank you again for your efforts and for sharing this information!

ZelnickB:
@idrassi Do you have a progress update on this feature?

josephazrak:
[...]

PR #1276, opened 2023-12-10
Improve and simplify macOS builds

denizt:
This series of commits aims to simplify the building of Veracrypt on macOS. This PR is currently draft and some commits require a rewrite as they are not atomic and require some fixes to the logic and the text.

1. Commit eca3086a1dac6af8d08a6d7b2ee9d9095dba2377 needs to be rewritten with the _packages_ location that Veracrypt internally uses, the public version in this commit is outdated
2. We need to make sure it's clear in the docs that homebrew instructions are only for development purposes

### DIFF
--- a/src/Main/Main.make
+++ b/src/Main/Main.make
@@ -210,7 +210,7 @@ else
 	sed -e 's/_VERSION_/$(patsubst %a,%.1,$(patsubst %b,%.2,$(TC_VERSION)))/' ../Build/Resources/MacOSX/Info.plist.xml >$(APPNAME).app/Contents/Info.plist
 endif
 	chmod -R go-w $(APPNAME).app
-ifneq ($(LOCAL_DEVELOPMENT_BUILD),"TRUE")
+ifneq ($(LOCAL_DEVELOPMENT_BUILD),"true")
 	codesign -s "Developer ID Application: IDRIX (Z933746L2S)" --timestamp $(APPNAME).app
 endif
 
--- a/src/Main/Main.make
+++ b/src/Main/Main.make
@@ -210,7 +210,9 @@ else
 	sed -e 's/_VERSION_/$(patsubst %a,%.1,$(patsubst %b,%.2,$(TC_VERSION)))/' ../Build/Resources/MacOSX/Info.plist.xml >$(APPNAME).app/Contents/Info.plist
 endif
 	chmod -R go-w $(APPNAME).app
+ifneq ($(LOCAL_DEVELOPMENT_BUILD),"TRUE")
 	codesign -s "Developer ID Application: IDRIX (Z933746L2S)" --timestamp $(APPNAME).app
+endif
 
 install: prepare
 	cp -R $(APPNAME).app /Applications/.
--- a/src/Makefile
+++ b/src/Makefile
@@ -295,7 +295,7 @@ ifeq "$(shell uname -s)" "Darwin"
 	PLATFORM := MacOSX
 	APPNAME := VeraCrypt
 
-	export VC_OSX_TARGET ?= 10.9
+	export VC_OSX_TARGET ?= 12
 	export VC_OSX_SDK ?= $(VC_OSX_TARGET)
 
 	#check to see if XCode 3 path exists.Otherwise, use XCode 4 path
--- a/src/Makefile
+++ b/src/Makefile
@@ -352,8 +352,21 @@ ifeq "$(shell uname -s)" "Darwin"
 		S := $(C_CXX_FLAGS)
 		C_CXX_FLAGS = $(subst -MMD,,$(S))
 
-		C_CXX_FLAGS += -gfull -arch $(CPU_ARCH)
-		LFLAGS += -Wl,-dead_strip -arch $(CPU_ARCH)
+		# only build local arch in development builds
+		ifeq "$(LOCAL_DEVELOPMENT_BUILD)" "true"
+			ifeq "$(CPU_ARCH)" "arm64"
+				C_CXX_FLAGS += -gfull -arch $(CPU_ARCH)
+				LFLAGS += -Wl,-dead_strip -arch $(CPU_ARCH)
+			else
+				C_CXX_FLAGS += -gfull -arch x86_64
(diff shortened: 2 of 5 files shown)


src/Makefile
```diff
@@ -350,8 +352,8 @@ ifeq "$(shell uname -s)" "Darwin"
 		S := $(C_CXX_FLAGS)
 		C_CXX_FLAGS = $(subst -MMD,,$(S))
 
-		C_CXX_FLAGS += -gfull -arch x86_64
```
Jertzukka, line 353: At the beginning of Makefile CPU_ARCH is set to `x64` based on my uname -m of `x86_64`, later CPU_ARCH is set to `x86` based on CPU_ARCH being `x64`. This means that here it'll attempt to build with arch `x86` which my Monterey doesn't have. This will error out with `clang: error: invalid arch name '-arch x86'`

src/Volume/Volume.make
```diff
@@ -37,6 +37,7 @@ endif
 
 ifeq "$(ENABLE_WOLFCRYPT)" "0"
 ifeq "$(PLATFORM)" "MacOSX"
+ifeq "$(COMPILE_ASM)" "true"
```
Jertzukka, line 40: `COMPILE_ASM` variable needs to be also set somewhere else than only in `build_veracrypt_macosx.sh`, like in `Makefile`, or else normal make will not work properly as it is not defined without this build script.

denizt: Thanks for the review @Jertzukka! 

Good call and find. I addressed this in 3e5168f1a6ceb9808aee8435a9a96ed15841797f, moving the check `ifneq` instead `ifeq` so the undefined variable doesn't fail the build. I'll try follow up on the other issues you mentioned tomorrow.

src/Makefile
denizt: Thank you so much for this @Jertzukka! I hadn't tested it on x86 and thought it could fail. I have now tried it on x86 and fixed all the problems with local builds using brew packages. I would appreciate it if you could confirm it on your end 😃

src/Makefile
```diff
@@ -295,7 +295,7 @@ ifeq "$(shell uname -s)" "Darwin"
 	PLATFORM := MacOSX
 	APPNAME := VeraCrypt
 
-	export VC_OSX_TARGET ?= 10.7
+	export VC_OSX_TARGET ?= 10.9
 	export VC_OSX_SDK ?= $(VC_OSX_TARGET)
```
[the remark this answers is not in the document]
denizt: I fixed this one in a follow-up commit and bumped it to 12, to be on the side of err,  as that was the target in the main script.

denizt:
@idrassi Can you please share/upload the [packages build](https://github.com/packagesdev/packages) that is used during the build of Veracypt, I believe the one on the webpage and homebrew is outdated and hosted on free.fr (HTTP).

I would also appreciate it if you could review the PR and have comments so I can wrap this up this weekend 😃 

Thanks!

vit9696:
@denizt, given the build changes, perhaps you could also introduce a switch to use a fuse-t implementation instead of macFUSE (ref #1055).

src/Main/Main.make
```diff
@@ -210,7 +210,9 @@ else
 	sed -e 's/_VERSION_/$(patsubst %a,%.1,$(patsubst %b,%.2,$(TC_VERSION)))/' ../Build/Resources/MacOSX/Info.plist.xml >$(APPNAME).app/Contents/Info.plist
 endif
 	chmod -R go-w $(APPNAME).app
+ifneq ($(LOCAL_DEVELOPMENT_BUILD),"true")
```
Jertzukka, line 213:
```suggestion
ifneq ("$(LOCAL_DEVELOPMENT_BUILD)","true")
```
This needs to be quoted to work. Also I might've chosen something shorter like NOSIGN which makes it also clear what the flag does, just skips the package sign parts.

Jertzukka: If we do not sign, the installer name in the later part will have wrong name that does not exist:

https://github.com/veracrypt/VeraCrypt/blob/6e28375060e043e9039bac4d292ecbcc5e94b08d/src/Main/Main.make#L233C2-L233C2

So, for no signing path, the copied installer name should be `"$(BASE_DIR)/Setup/MacOSX/VeraCrypt $(TC_VERSION).pkg"` instead of `"$(BASE_DIR)/Setup/MacOSX/VeraCrypt_$(TC_VERSION).pkg"`

denizt: Thanks! I'll have a look at this (and possibly others) once the maintainer reviews

denizt:
> @denizt, given the build changes, perhaps you could also introduce a switch to use a fuse-t implementation instead of macFUSE (ref #1055).

I can do that in a follow up commit, this PR is unfortunately stalled and I would like to wrap this up first 😄 

@idrassi Can you please review?

idrassi:
Sorry for the delay! Things have been tough on my side lately. Although there are some minor issues with this PR, I'm merging it, and I will address the issues afterward. Here are some comments about src/Makefile:

- $(CPU_ARCH) is not needed, so I will remove it.
- The current logic always adds flags for the x64 build, and then we add either x86 or arm64 flags depending on whether we are doing a legacy or normal build, respectively. In this PR, changes were made to both checks. A better approach would be to have a dedicated block for the development build where we use only native flags (either x64 or arm64) and an official build block containing the existing logic verbatim. This will be much clearer and simpler for maintenance.

idrassi:
> @idrassi Can you please share/upload the [packages build](https://github.com/packagesdev/packages) that is used during the build of Veracypt, I believe the one on the webpage and homebrew is outdated and hosted on free.fr (HTTP).
> 
> I would also appreciate it if you could review the PR and have comments so I can wrap this up this weekend 😃
> 
> Thanks!

@denizt I have uploaded the notarized build of Packages that I use for official VeraCrypt build to https://github.com/idrassi/packages/releases. I have also updated the Readme to point to this link.

After merging this PR, I have introduced some enhancements to your changes. You can check and share your comments.

Next step is FUSE-T support (https://github.com/veracrypt/VeraCrypt/issues/1055)

PR #1288, opened 2024-01-05
Unix CLI: Don't initially re-ask PIM if it was already specified

### LINKED ISSUE #1286: pim=0 no longer works
On mount pim=0 option doesn't work anymore. It asks to enter something or leave blank and click enter.

veracrypt -t "$fName" "$mName" --pim=0 -p "" -k "" --protect-hidden=no --fs-options=noatime

VeraCrypt 1.26.7 + Linux x64

# Expected behavior
Don't ask anything, the only question should be about password, not pim, since pim is disabled in command line

# Observed behavior
Asks to enter pim. You need to press Enter

**VeraCrypt version:** VeraCrypt 1.26.7

**Operating system and version:** Ubuntu 22.04

**System type:** 64-bit

Jertzukka:
We don't need to reset PIM in PasswordException as it is immediately fell back to if PIM is specified in text mode, but password is not. This causes an exception that resets the PIM when it shouldn't. If a password is given incorrectly, we still get caught by PasswordIncorrect and PIM is correctly reset to -1 as proposed and implemented in https://github.com/veracrypt/VeraCrypt/commit/eb61010ce2f1e9ed16f136500c6e76d2839bfd08 based on #1026.

Fixes #1286

### DIFF
--- a/src/Main/TextUserInterface.cpp
+++ b/src/Main/TextUserInterface.cpp
@@ -1409,7 +1409,6 @@ namespace VeraCrypt
 			{
 				ShowInfo (e);
 				options.Password.reset();
-				options.Pim = -1;
 			}
 		}
 


idrassi:
Thank you for fixing this.

PR #1292, opened 2024-01-10
Allow compilation with g++17 by replacing "byte" with "uint8_t" to remove confusion with std::byte (c++17 feature)

segfaultDelirium:
C++17 features std::byte which breaks "byte" definitions in Crypto++ library. This pull requests fixes this.


idrassi:
I have implemented a similar change using the uint8 type as I mentioned earlier: https://github.com/veracrypt/VeraCrypt/commit/455a4f2176a5cfbe325e1e40cea20dd3e466b64c
So closing this PR.

PR #1299, opened 2024-02-02
Fix exceptions to match the correct language strings

Jertzukka:
This is something I accidentally stumbled upon while searching for something else, and thought it looked incorrect when the exception name didn't match the language string it's supposed to be associated with.

Fixes the order of the language strings to match the correct exceptions.


idrassi:
Thank you for fixing this mistake, I appreciate.

PR #1301, opened 2024-02-07
07Feb23-Update Language.es.xml

MayanTigger:


idrassi:
Thanks. This seems to be a better translation than previous PR.
Only issue  is the minor punctuation error (double period) in the "LINUX_EX2MSG_KERNELCRYPTOSERVICETESTFAILED" entry, which should be corrected to a single period this is something I can do on my side.

PR #1306, opened 2024-02-16
Updated translation: pt-br.

nerun:
Translated some messages in pt-br.


idrassi:
"PIM" acronym should remain unchanged in the Portuguese translation to maintain consistency with the original acronym's recognition and usage. Si it should not be replaced by "MIP". Other translations keep "PIM" acronym unchanged.

@nerun  Can you please update the translation to keep "PIM" acronym unchanged?

Also, in the translation for "IDT_SYS_PARTITION," there's an extra period at the end of the sentence in Portuguese ("...está instalado.."). It should be corrected to a single period to match the original text.

Other than this, other parts look ok. Thank you for your contribution.

nerun:
Thanks for the review. Changes made.

idrassi:
The new translation of IDT_FILE_SYS should be "Sistema de arquivos: " (in plural form) like in the translation of IDT_FILESYSTEM entry in this same file. This will ensure consistency across all entries.
Can you please do this change?

nerun:
Done, thank you!

idrassi:
Thank you for the quick change. All is good now!

PR #1338, opened 2024-05-08
Fix incorrect max hidden volume size for file containers on CLI

Jertzukka:
Currently the maximum hidden volume size for file containers is limited by available free space on the device the file container resides on, which we do not care about. This commit changes so that only Normal volumes get their `maxVolumeSize` limited by `AvailableDiskSpace` as the file that is going to host the Hidden volume is already confirmed to exist. Also the --size=max parameter is restricted from hidden volume creation as there is no way to determine a good size as we do not mount the outer volume through the CLI process flow to determine available free space on the outer volume.

Tested on Linux with multiple device/file container, size and parameter options, and shouldn't cause any regressions.

### DIFF
--- a/src/Main/TextUserInterface.cpp
+++ b/src/Main/TextUserInterface.cpp
@@ -668,7 +668,7 @@ namespace VeraCrypt
 				{
 					parentDir = wxT(".");
 				}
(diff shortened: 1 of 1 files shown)


idrassi:
Thank you for this enhancement. I agree with your proposed approach to force the explicit selection of hidden volume size.
Merging your PR.

Issue #21, opened 2015-08-14
How can I decrypt Vera-crypted 'system' drive on other computer?


espress0:
PC1: Encrypted entire 'System Drive' with VC

PC1's windows have problem, it goes straight to BSOD when booting.
So I took a HDD from PC1.
I want to decrypt it on PC2(and I know a password).

How can I do it?

Must:
1. I don't want to pull out PC2's HDD.
2. I have a HDD-to-USB cable set.
3. VC installed on PC2.

idrassi:
Currently, only the rescue disk can be used to decrypt an encrypted system drive.
Decrypting system drive on another machine without a rescue disk is not supported currently.

It is planned to implement this feature although the original bootloader can not be restored since it is present in the rescue disk. The user will have to use a Windows installation disk to restore Windows bootloader.

espress0:
Why can't I just remove HDD deom PC1 and decrypt it on PC2(running Windows & latest veracrypt)?

I really need to decrypt it now because of bad windows update. PC1 goes to unlimited loop, safe mode doesn't work, each time I have to imput password...

Please add it. I need it.

espress0:
And I didn't have rescue disk. But I know the password.

espress0:
> The user will have to use a Windows installation disk to restore Windows bootloader.

I already have Windows 7 DVD. So it's OK.
I just want to decrypt the disk, insert Windows DVD to fix bootloader, and do something to fix this mess!

espress0:
In recent days Microsoft send PCs untested/unstable updates. I'm the victim of M$.

it-solutions-zehaczek:
I just had to decrypt my sys using the rescue disc because of a broken windows and CloneZilla for some reason cannot restore my VeraCrypt drive, the Boot Loader passes however windows cannot found its own system later.

The Rescue Disc says this is really slow and YES IT IS. I plugged the drive onto another computer and can successfully mount, but not decrypt the partition. I would be so happy if there will be a permanently decrypt of system partition not in place in a future release.

formtapez:
you can use veracrypt on a live linux system to mount your system partition (sda4 in my case), then "dd" the decrypted virtual blockdevice to a safe place, and write it back to disk:

- boot a live linux system
- install veracrypt
- mount system partition using veracrypt (read-only!)
- dd if=/dev/mapper/veracrypt1 of=/mnt/somewhere/decrypted.img bs=64M status=progress
[...]

Issue #20, opened 2015-08-14
Add "Create language files and PDF guide" to installer.


espress0:
To me, these files are junk. I deleted everytime I upgrade Veracrypt.

Language.*.xml (I need English only)
LICENSE
License.txt (Why do you need it? I can read it anytime from Help->About)
NOTICE (I'll read readme.txt from sourceforge)
Veracrypt user guide.pdf (I already copied it to somewhere else, no need to create again)

idrassi:
Concerning the language files, I understand your point but please use a better language: calling other languages files junk is pretty harsh and insensitive to non English speakers.
We understand that you need English only but VeraCrypt approach was to enable an international support by default to minimize development efforts. Indeed, adding such option to the installer will be good, but your words to explain it are aggressive and out of touch.

The three other files are needed for licensing pupose and can't be removed: LICENSE and NOTICE are required by the Apache 2.0 license. License.txt is required by TrueCrypt License 3.0.

The VeraCrypt user guide is embedded so that users don't need to connect to the internet to get one. For this one, I don't agree about removing it.

In general, please try to be more sensitive in your writing and avoid expressing your wishes like orders. This is an open source software and all ideas are welcomed by they must be expressed in a good way. No body is taking orders from anybody here.

andreas-becker:
@idrassi, seems like this issue can be closed. The user is not on github anymore.

Issue #19, opened 2015-08-14
v1.1f -> v1.13 after reboot, this error pop up.


espress0:
---------------------------
VeraCrypt
---------------------------
The system cannot find the path specified.

Source: MainDialogProc:5874
---------------------------
OK   
---------------------------

idrassi:
Obviously, no one can help you without giving more details.

This error happens when VeraCrypt can not find the file that you specified in its command line.
Here is the code in question: https://github.com/veracrypt/VeraCrypt/blob/master/src/Mount/Mount.c#L5875

Basically, a file name / device name was specified in the command line and VeraCrypt can't find it.

I suspect that you are using a script to mount volumes at logon and in this script you are using an unsupported syntax were the parameters can be placed directly after their switch (for example /le instead of /l e).

Can you please post the command line you are using to mount volumes are boot?

In general, try to give more information to help solve your issue. Just posting an error message without any context is not helpful at all.

Since in the title you are talking about reboot, I guess you have some favorite configured to be mounted at logon and VeraCrypt is unable to find the specified file container or partition.

Please, if you want us to help, give more details: are you using system encryption or not? Do you favorites configured? What type of favorites do you have? Did you check if the volumes are available? etc...

espress0:
> device name was specified in the command line and VeraCrypt can't find it

My bad. Looks like the volume serial was changed when I reconnect my HDD.

PR #1343, opened 2024-05-15
Prepare for changes in wxWidgets 3.3

Jertzukka:
wxWidgets 3.3 which is going to be released this autumn according to the roadmap, is going to introduce some ABI breakage. This PR is going to suggest changes that ensure that when 3.3 hits the distro repositories, VeraCrypt will build against it.

Main breakage is caused by our reliance on implicit conversions from wxString, and as of [this commit](https://github.com/wxWidgets/wxWidgets/commit/35c35c235e9c29b40002131602e050dca8d65b8c) and the commit mentioned in it's message, wxWidgets 3.3 defaults to using STL classes which require explicit conversions, so our implicit conversions won't build anymore.

Another change in 3.3 caused the header for wxScopedPtr not being included anymore through our other includes. wxScopedPtr is deprecated and upstream recommends using STL smart pointers instead, so instead of re-including the header for the deprecated function, we can just change it to std::unique_ptr and avoid the include.

These changes mainly increase verbosity, explicitness and won't affect building against 3.2 which VeraCrypt currently uses. These changes make building against current wxWidgets master branch possible (with few small Makefile changes).

### DIFF
--- a/src/Main/Application.cpp
+++ b/src/Main/Application.cpp
@@ -89,7 +89,7 @@ namespace VeraCrypt
 
 	FilePath Application::GetConfigFilePath (const wxString &configFileName, bool createConfigDir)
 	{
-		static wxScopedPtr<const wxString> configDirC;
+		static std::unique_ptr<const wxString> configDirC;
 		static bool configDirExists = false;
 
 		if (!configDirExists)
--- a/src/Main/Forms/VolumeCreationWizard.cpp
+++ b/src/Main/Forms/VolumeCreationWizard.cpp
@@ -975,7 +975,7 @@ namespace VeraCrypt
 						if (OuterVolume && VolumeSize > TC_MAX_FAT_SECTOR_COUNT * SectorSize)
 						{
 							uint64 limit = TC_MAX_FAT_SECTOR_COUNT * SectorSize / BYTES_PER_TB;
-							wstring err = StringFormatter (LangString["LINUX_ERROR_SIZE_HIDDEN_VOL"], limit, limit * 1024);
+							wstring err = static_cast<wstring>(StringFormatter (LangString["LINUX_ERROR_SIZE_HIDDEN_VOL"], limit, limit * 1024));
 
 							if (SectorSize < 4096)
 							{
--- a/src/Main/Resources.cpp
+++ b/src/Main/Resources.cpp
@@ -71,12 +71,12 @@ namespace VeraCrypt
 
 		UserPreferences Preferences;
 		Preferences.Load();
-		wstring preferredLang = Preferences.Language;
+		string preferredLang = string(Preferences.Language.begin(), Preferences.Language.end());
 #ifdef DEBUG
 		std::cout << "Config language: " << preferredLang << std::endl;
 #endif
 
-		if (preferredLang == L"system") {
+		if (preferredLang == "system") {
 			if (const char *env_p = getenv("LANG")) {
 				string lang(env_p);
 #ifdef DEBUG
--- a/src/Main/StringFormatter.h
+++ b/src/Main/StringFormatter.h
@@ -52,7 +52,7 @@ namespace VeraCrypt
 		StringFormatter (const wxString &format, StringFormatterArg arg0 = StringFormatterArg(), StringFormatterArg arg1 = StringFormatterArg(), StringFormatterArg arg2 = StringFormatterArg(), StringFormatterArg arg3 = StringFormatterArg(), StringFormatterArg arg4 = StringFormatterArg(), StringFormatterArg arg5 = StringFormatterArg(), StringFormatterArg arg6 = StringFormatterArg(), StringFormatterArg arg7 = StringFormatterArg(), StringFormatterArg arg8 = StringFormatterArg(), StringFormatterArg arg9 = StringFormatterArg());
 		virtual ~StringFormatter ();
 
-		operator wstring () const { return wstring (FormattedString); }
+		explicit operator wstring () const { return wstring (FormattedString); }
 		operator wxString () const { return FormattedString; }
 		operator StringFormatterArg () const { return FormattedString; }
 
--- a/src/Main/UserInterface.cpp
+++ b/src/Main/UserInterface.cpp
@@ -390,7 +390,7 @@ namespace VeraCrypt
 			errOutput += StringConverter::ToWide (execEx->GetErrorOutput());
 
 			if (errOutput.empty())
-				return errOutput + StringFormatter (LangString["LINUX_COMMAND_GET_ERROR"], execEx->GetCommand(), execEx->GetExitCode());
+				return errOutput + static_cast<wstring>(StringFormatter (LangString["LINUX_COMMAND_GET_ERROR"], execEx->GetCommand(), execEx->GetExitCode()));
 
 			return wxString (errOutput).Trim (true);
 		}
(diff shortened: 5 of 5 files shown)


idrassi:
Thank you for these proactive changes. I'm merging them.

can please share what kind of Makefile changes will be needed?

Jertzukka:
> Thank you for these proactive changes. I'm merging them.
> 
> can please share what kind of Makefile changes will be needed?

https://github.com/veracrypt/VeraCrypt/blob/114624b3a57374d69e4e4c7c2dd9d9f8ac308489/src/Makefile#L490

Parameters --enable-unicode (see https://github.com/wxWidgets/wxWidgets/commit/a60c9b0f5e934d588dd91e73c4db4c8335f664f0) and --enable-std_string (see https://github.com/wxWidgets/wxWidgets/commit/be7860c7661756e8de36985ade760d5c9a784ab7) are removed, which makes wxbuild fail with these options. If I understand the bash mess in the wxWidgets configure correctly, both of these values were already enabled by default in 3.2.2.1, so I don't think removing these options would affect our current build configuration. You should be safe to remove them.

idrassi:
Thank you for sharing these details, very clear now.

By the way, I'm battling a crash of VeraCrypt GUI build in text mode (--text) when building and running under Ubuntu 24.04 (which ships version 3.2.4). The binary by the command "make" crashes when run with "veracrypt --text --test". gdb shows this:

<pre>#0  <font color="#A2734C">__pthread_kill_implementation</font> (no_tid=0, signo=6, threadid=&lt;optimized out&gt;) at <font color="#26A269">./nptl/pthread_kill.c</font>:44
#1  <font color="#A2734C">__pthread_kill_internal</font> (signo=6, threadid=&lt;optimized out&gt;) at <font color="#26A269">./nptl/pthread_kill.c</font>:78
#2  <font color="#A2734C">__GI___pthread_kill</font> (threadid=&lt;optimized out&gt;, signo=signo@entry=6) at <font color="#26A269">./nptl/pthread_kill.c</font>:89
#3  <font color="#12488B">0x00007ffff6a4526e</font> in <font color="#A2734C">__GI_raise</font> (sig=sig@entry=6) at <font color="#26A269">../sysdeps/posix/raise.c</font>:26
#4  <font color="#12488B">0x00007ffff6a288ff</font> in <font color="#A2734C">__GI_abort</font> () at <font color="#26A269">./stdlib/abort.c</font>:79
#5  <font color="#12488B">0x00007ffff6a297b6</font> in <font color="#A2734C">__libc_message_impl</font> (fmt=fmt@entry=0x7ffff6bce8d7 &quot;%s\n&quot;) at <font color="#26A269">../sysdeps/posix/libc_fatal.c</font>:132
#6  <font color="#12488B">0x00007ffff6aa8fe5</font> in <font color="#A2734C">malloc_printerr</font> (str=str@entry=0x7ffff6bd1ac0 &quot;double free or corruption (out)&quot;)
    at <font color="#26A269">./malloc/malloc.c</font>:5772
#7  <font color="#12488B">0x00007ffff6aab110</font> in <font color="#A2734C">_int_free_merge_chunk</font> (av=0x7ffff6c03ac0 &lt;main_arena&gt;, p=0x7fffffffd620, size=140736951495520)
    at <font color="#26A269">./malloc/malloc.c</font>:4676
#8  <font color="#12488B">0x00007ffff6aab42a</font> in <font color="#A2734C">_int_free</font> (av=0x7ffff6c03ac0 &lt;main_arena&gt;, p=&lt;optimized out&gt;, have_lock=&lt;optimized out&gt;)
    at <font color="#26A269">./malloc/malloc.c</font>:4646
#9  <font color="#12488B">0x00007ffff6aadd9e</font> in <font color="#A2734C">__GI___libc_free</font> (mem=0x7fffffffd630) at <font color="#26A269">./malloc/malloc.c</font>:3398
#10 <font color="#12488B">0x00007ffff796bfb9</font> in <font color="#A2734C">UpdatePreferDark</font> (value=value@entry=0x7fffe0002730) at <font color="#26A269">./src/gtk/settings.cpp</font>:202
#11 <font color="#12488B">0x00007ffff796c136</font> in <font color="#A2734C">wxSystemSettingsModule::OnInit</font> (this=&lt;optimized out&gt;) at <font color="#26A269">./src/gtk/settings.cpp</font>:1240
#12 <font color="#A2734C">wxSystemSettingsModule::OnInit</font> (this=&lt;optimized out&gt;) at <font color="#26A269">./src/gtk/settings.cpp</font>:1200
#13 <font color="#12488B">0x00007ffff730fe76</font> in <font color="#A2734C">wxModule::Init</font> (this=0x555555b6b410) at <font color="#26A269">./include/wx/module.h</font>:33
#14 <font color="#A2734C">wxModule::DoInitializeModule</font> (module=0x555555b6b410, initializedModules=...) at <font color="#26A269">./src/common/module.cpp</font>:137
#15 <font color="#12488B">0x00007ffff73106b8</font> in <font color="#A2734C">wxModule::InitializeModules</font> () at <font color="#26A269">./src/common/module.cpp</font>:168
#16 <font color="#12488B">0x00007ffff72f89dd</font> in <font color="#A2734C">DoCommonPostInit</font> () at <font color="#26A269">./src/common/init.cpp</font>:270
#17 <font color="#12488B">0x00007ffff72f9195</font> in <font color="#A2734C">wxEntryStart</font> (argc=&lt;optimized out&gt;, argv=&lt;optimized out&gt;) at <font color="#26A269">./src/common/init.cpp</font>:336
#18 <font color="#12488B">0x00007ffff72f9365</font> in <font color="#A2734C">wxInitialize</font> (argc=&lt;optimized out&gt;, argv=&lt;optimized out&gt;) at <font color="#26A269">./src/common/init.cpp</font>:532
#19 <font color="#12488B">0x00007ffff72f94a0</font> in <font color="#A2734C">wxInitializer::wxInitializer</font> (argv=&lt;optimized out&gt;, argc=&lt;optimized out&gt;, this=&lt;synthetic pointer&gt;)
    at <font color="#26A269">./include/wx/init.h</font>:92
#20 <font color="#A2734C">wxEntry</font> (argc=&lt;optimized out&gt;, argv=&lt;optimized out&gt;) at <font color="#26A269">./src/common/init.cpp</font>:467
#21 <font color="#12488B">0x0000555555685e09</font> in <font color="#A2734C">main</font> (argc=3, argv=0x7fffffffde18) at <font color="#26A269">Unix/Main.cpp</font>:107
</pre>

The crash is caused by a double-free inside wxWidgets code but I cannot find the reason. Previous versions of wxWidgets under Ubuntu 23.10 and old didn't cause any issues in text mode for GUI binary.

Any help is more than welcomed.

Jertzukka:
@idrassi I can replicate your issue with 3.2.4 from Ubuntu 24.04 repos and the tarball downloaded from their original 3.2.4 Github release. Though if I build against latest stable, wxWidgets 3.2.5, I'm having no issues. In 3.2.5 they introduced a patch `Fix crash with GTK3 if console program is using a GUI wxApp class (#24081).` which references https://github.com/wxWidgets/wxWidgets/issues/24081. This commit https://github.com/wxWidgets/wxWidgets/commit/8ea22b5e92bf46add0b20059f6e39a938858ff97 most likely fixed the crash issue in text mode.

The 3.2.5 release is only a week old, so it might take a while until Ubuntu has it in their repositories. Meanwhile I'd recommend building any new binaries against a self-compiled 3.2.5, and as the 3.2 branch is ABI compatible with eachother, there shouldn't be any problems (?)

I can also see a problem with my own testing and how I didn't notice this issue earlier as the tests are only run in Release mode and I always compile with `DEBUG=1`.

idrassi:
Thanks @Jertzukka for the quick feedback. Amazing find!

I agree that using self-compiled wxWidgets is the way to go for Ubuntu 24.04. I will give it a try.

PR #1349, opened 2024-05-29
FreeBSD: Fix privilege escalation prompts not showing up

Jertzukka:
The behaviour of `wc -l` is different on FreeBSD, in which the stdout result is padded by spaces in the beginning, which causes that the `result[0]` is not actually the value we care about. This patch adds a translate removing all whitespace from the output.


idrassi:
Thank you for the fix.

Issue #576, opened 2020-01-14
VeraCrypt on FreeBSD: "Error: mount: /dev/md0: No such file or directory"


strfry:
This has already been posted, but the sourceforge bugtracker doesn't seem to be getting much attention:
https://sourceforge.net/p/veracrypt/tickets/263/

As the description implies, handling of memory disks has changed on recent FreeBSD versions, which hasn't been corrected in Veracrypt yet.
I haven't tested if this actually works with FAT partitions, as described by the original reporter.

Still, i need this for work, and would be willing to look into it.
I would appreciate some guidance though, where to look for the thing that went wrong here.

strfry:
I found an answer to this problem.
Apparently, the FreeBSD backend doesn't even bother about detecting any fs types than FAT.
And even if it did, `mount -t ntfs` would not work, as ntfs read support has been removed from kernel, and we're left with ntfs-3g, which has so be specified via `-o mountprog=/usr/local/bin/ntfs-3g`.

Which works for me as a hardcoded fix (i don't actually need FAT at all in Veracrypt), but it would be nice to have a general solution.
Since coding a filesystem is a bit of a task, i wonder if there is another way to influence what is passed as `filesystemType` to `CoreFreeBSD::MountFileSystem`?

I found Veracrypt a bit hard to debug and follow because of the forking multi-process architecture. So a little advise would be appreciated, on to make a proper fix for this issue.

andreas-becker:
@idrassi maybe you can give @strfry a little advise where to fix the issue so he can create a pull request.

skateboardCherimoya:
I also need to mount ntfs devices on FreeBSD and would be very glad if this issue is resolved in the future.
@strfry Can you please tell me what exactly you patched in order to get this to work?

o-alquimista:
@idrassi Reopen please. Still reproducible.

https://sourceforge.net/p/veracrypt/discussion/technical/thread/667da9e7e7/?limit=25
https://sourceforge.net/p/veracrypt/tickets/263/

I'm trying to mount an ext3 partition instead.

Is there any chance this would work on OpenBSD? Did anyone try? I'm about to.

idrassi:
Indeed, it seems the automatic fs detection doesn't work anymore in FreeBSD.
I'm not sure about mdconfig link since VeraCrypt volume requires on-the-fly encryption/decryption while mdconfig is just for RAM disks. This seems to be more related to how filesystem is detected rather than the virtual device itself.
[...]

PR #1350, opened 2024-05-29
FreeBSD: Support automatic detection and mounting of ext2/3/4, exFAT, NTFS filesystems

Jertzukka:
Currently the `mount` implementation on FreeBSD does not support automatic filesystem detection on mount. We can alternatively use `blkid` which allows us to detect the filesystem that is contained on the virtual device before we attempt to mount the filesystem. The `blkid` program is part of the `e2fsprogs` package, which also provides the ext2/3/4 driver and is very likely to be installed on the user's system, if not, it will silently fail and behave like it has previously.

This patch adds support for automatic mounting of:
- Ext2/3/4 (requires `e2fsprogs`)
- exFAT via mount.exfat (requires `fusefs-exfat`)
- NTFS via ntfs-3g (requires `fusefs-ntfs`)

Related: https://github.com/veracrypt/VeraCrypt/issues/576

### DIFF
--- a/src/Core/Unix/FreeBSD/CoreFreeBSD.cpp
+++ b/src/Core/Unix/FreeBSD/CoreFreeBSD.cpp
@@ -185,10 +185,41 @@ namespace VeraCrypt
 
 	void CoreFreeBSD::MountFilesystem (const DevicePath &devicePath, const DirectoryPath &mountPoint, const string &filesystemType, bool readOnly, const string &systemMountOptions) const
 	{
+		std::string chosenFilesystem = "msdos";
+
+		// No filesystem type specified through CLI, attempt to identify with blkid
+		// as mount is unable to probe filesystem type on BSD
+		if (filesystemType.empty()) {
+			std::vector<char> buffer(128,0);
+			std::string result;
+			std::string command = std::string("blkid -o value -s TYPE ") + static_cast<std::string>(devicePath);
+
+			FILE* pipe = popen(command.c_str(), "r");
+			if (pipe) {
+				while (!feof(pipe)) {
+					if (fgets(buffer.data(), 128, pipe) != nullptr)
+						result += buffer.data();
+				}
+				fflush(pipe);
+				pclose(pipe);
+				pipe = nullptr;
+
+				if (!result.empty() && strlen(result.c_str()) != 0) {
+					if (result.find("ext") == 0)
+						chosenFilesystem = "ext2fs";
+				}
+			}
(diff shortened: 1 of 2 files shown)


src/Core/Unix/FreeBSD/CoreFreeBSD.cpp
```diff
@@ -185,17 +185,58 @@ namespace VeraCrypt
...
-			CoreUnix::MountFilesystem (devicePath, mountPoint, filesystemType.empty() ? "msdos" : filesystemType, readOnly, systemMountOptions);
+			CoreUnix::MountFilesystem (devicePath, mountPoint, chosenFilesystem, readOnly, modifiedMountOptions);
 		}
 		catch (ExecutedProcessFailed&)
 		{
 			if (!filesystemType.empty())
 				throw;
 
-			CoreUnix::MountFilesystem (devicePath, mountPoint, filesystemType, readOnly, systemMountOptions);
+			CoreUnix::MountFilesystem (devicePath, mountPoint, filesystemType, readOnly, modifiedMountOptions);
```
idrassi, line 239: @Jertzukka :
Is there a specific reason why `modifiedMountOptions `is retained in the catch block instead of reverting to `systemMountOptions`? `modifiedMountOptions `is directly linked to `chosenFilesystem `so using it with `filesystemType  `doesn't look right

idrassi (commented):
query about src/Core/Unix/FreeBSD/CoreFreeBSD.cpp changes posted.

src/Core/Unix/FreeBSD/CoreFreeBSD.cpp
Jertzukka: Only paths where `modifiedMountOptions` changes is if:

- blkid supplies exfat or ntfs result, in this case using `systemMountOptions` which omits our specified mountprog will never successfully mount.
- User specifies `--filesystem=exfat` or `--filesystem=ntfs` through CLI, let's say the volume is actually UFS. First mount fails, we throw because `if (!filesystemType.empty()) throw;` and we never get to the second mount prompt.

I remember thinking this through and coming to a conclusion that in practice using either `systemMountOptions` and `modifiedMountOptions` here has no difference, as the only edge case where it matters would be where blkid misidentifies a volume as ntfs/exfat where it was actually something like UFS. But I agree, it is probably clearer for anyone reading the code not having to do 10 steps of logic to see why a variable that refers to something `modified` is used here in a fallback path.

idrassi:
Thank you for the quick update.

PR #1354, opened 2024-06-03
Avoid assert by verifying folder for installed languages exists

Jertzukka:
wxDir::GetAllFiles will throw an assert when opening the preferences dialog if for some reason the user has not installed VeraCrypt properly and is missing the intended folder. This patch adds a check to ensure the folder first exists before querying its files.

There was two options, leaving the two options of "System language" and "English" in the combo box when no languages are found, or putting these two options also inside the `if` block leaving the combo box empty. I think it's clearer to the user when they only see two options to wonder "how do I get more languages?" instead of being faced with an empty selection box and wondering if it is bugged.


idrassi:
Thank you for the detailed explanation. I agree with your approach.

PR #1360, opened 2024-06-17
Improve documentation

denizt:
This PR/commit brings the following improvements to the README.md

- Ran a spell check and correct mistakes
- Markdown markers are used so you can navigate in the file
- Used markdown local links for documentation
- Removed mention of now-outdated packages since the notarized packages is published and preferred.

And other small improvements.


idrassi:
Thank you for these enhancements

PR #1367, opened 2024-06-23
Fix regression in core service deserialization

Jertzukka:
Introducing data into the standard out in the core service will cause deserialization problems, as the pipes are used for interprocess communication.

I don't even think you would see this message anywhere on the terminal as the core service is a forked process. If you wanted to print or log the errors when attempting to mount FUSE devices, you could do something like:

`SystemLog::WriteError("Exception. Error mounting volume: " + static_cast<string>(*options.Path));`

Which would cause it to be printed in either syslog on Linux, or to the Console logs in macOS.

### DIFF
--- a/src/Core/Unix/CoreUnix.cpp
+++ b/src/Core/Unix/CoreUnix.cpp
@@ -728,7 +728,6 @@ namespace VeraCrypt
 		}
 		catch (...)
 		{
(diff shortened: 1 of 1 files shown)


idrassi:
Thank you for catching this! This is a leftover from my debugging. I thought I had cleaned up everything but clearly I forgot this one!

PR #1371, opened 2024-07-01
MacOS: Support CommandLineTools for building

Jertzukka:
Full Xcode application which can take up-to 40GB of disk space is not necessary for building VeraCrypt, rather the CommandLineTools for XCode are sufficient which only take few gigabytes. Instead of hardcoding the SDK location, use xcrun --show-sdk-path to support also the CommandLineTools instance.

The reason for switching the logic for the XCode version is because the xcodebuild will not report a correct XCode version for the CommandLineTools, thus it can't be relied for whether to use the -Wl,-ld_classic flags. Instead, we can just check the ld version in use in the active developer directory, and see whether we are using dyld (the new) or ld64 (the old).


idrassi:
Thank you for these changes. CommandLineTools support is indeed helpful, and your new method to check for the new linker is clever!
Nice contribution as usual.

PR #1378, opened 2024-07-16
allow noattach on commandline

bernd-elk:


bernd-elk:
noattach works for me on freebsd (with my workaround to prevent the errormessage), it seems that the code is used not-only on windows.
I did not know about --filesystem=none, did not even look there only in the mount options. Will be great if it does the same. I'll close the PR.

PR #1379, opened 2024-07-19
Updated pt-br translation.

nerun:
Updated a few entries.

In a few cases, some already translated entries were edited, not because they were wrong, but to better fit the space for which they were intended.


idrassi:

PR #1382, opened 2024-08-01
Add Norwegian Bokmål translation

sandakersmann:
Add Norwegian Bokmål translation

### DIFF
--- a/Translations/Language.nb.xml
+++ b/Translations/Language.nb.xml
@@ -1,0 +1,1684 @@
+<?xml version="1.0" encoding="utf-8"?>
+<VeraCrypt>
+  <localization prog-version= "1.26.12">
+    <language langid="nb" name="Norsk Bokmål" en-name="Norwegian (Bokmål)" version="0.1.0" translators="Marius Kjærstad" />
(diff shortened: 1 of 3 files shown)


sandakersmann:
@idrassi This is translated with [https://nano-gpt.com](https://nano-gpt.com) GPT 4o. So some fixes needed, but I think it's a good start :)

idrassi:
Thank you for this contribution. This is a good use of AI. It must have taken you a lot of prompts to achieve this result.
For sure, we need a human feedback/review but it is better than not having any translation.

lollolong:
Hi @sandakersmann & @idrassi , this PR seems to break windows compilation because the resource entry for `IDR_LANG_NB` isn't properly set up.
`error C2065: 'IDR_LANG_NB' : undeclared identifier`

sandakersmann:
I will fix it.

sandakersmann:
This should fix the issue: https://github.com/veracrypt/VeraCrypt/pull/1388

idrassi:
Thank you @lollolong for spotting this.
@sandakersmann PR has been merged.

PR #1387, opened 2024-08-03
Update Language.nl.xml

RockyTDR:
Update Dutch translations


idrassi:
Thank you for the translation.

PR #1389, opened 2024-08-05
Windows: disable dropdown split button in volume expander

lollolong:
This change removes the split button in the Volume Expander because the "Mount Without Cache" option isn't available. Having a dropdown button without a dropdown menu behind it can be confusing.

image

image


idrassi:
Thank you for fixing the button visual.

PR #1394, opened 2024-08-09
Update Language.sv.xml

JonatanWick:
Update Swedish translation


idrassi:
I have one query: I started to use AI for doing quick check of submitted translations and it has two remarks about your translation:
- ERR_XTS_MASTERKEY_VULNERABLE: Adding "Vänligen" (Please) makes it more consistent with the English version and adds a polite tone.
- ERR_XTS_MASTERKEY_VULNERABLE_SHORT: "Säkerhetssårbarhet" more accurately translates "security vulnerability" than "säkerhetsrisk" (security risk).

I'm interested in knowing your feedback about these suggestions.
Thanks.

JonatanWick:
"Please" is usually not translated to "Vänligen" in Swedish but is left out but I don't know the reason behind that. I will ask around what is the reason as I am also interested to know. You are right about "Säkerhetssårbarhet", I have opened a new pr for that.

idrassi:
Thank you for the feedback and the new PR. It is interesting to see human nuances not caught by AI.

JonatanWick:
I got the answer from another translator that it is mainly because English has a much higher use of "please" than the equivalent is used in Swedish. I will continue in that style for new translations.

PR #1400, opened 2024-08-13
Update Finnish translation and few xml errors

Jertzukka:
Few updates on the Finnish translation and few errors in the translation XML's found via https://github.com/Jertzukka/VeraCrypt/blob/ci/.github/workflows/xmlvalidate.yml and https://github.com/Jertzukka/VeraCrypt/blob/ci/.github/workflows/xmlvalidate.sh. 

If it would be helpful for you, I could fine tune it and PR it for this repository to ensure that translation PR's don't have errors in the future. It currently runs all translation files through fast-xml-parser and also ensures that all translation files have all the keys that exist in the common Language.xml.


idrassi:
Thank you for the translation update and for fixing the XML issue.
Indeed, it would be helpful to add your validation scripts to VeraCrypt workflows. You contribution to this will be much appreciated.

By the way, I have sent you a message using Sourceforge contact form few weeks ago, not sure if you read it. Things I mentioned in my message are to be taken with a grain of salt as I'm not sure anymore. To good to be true...

PR #1401, opened 2024-08-15
Added some translations to Thai

sandakersmann:
Translated with ChatGPT-4o


idrassi:
Thank you for the translations. I cannot review all entries but I didn't some checks and their seem to be possible enhancements and fixes.
My method is to combine both ChatGPT-4o and Gemini 1.5 Pro Experimental (through Google AI Studio) using multi-round chats. This is time consuming but it reduces the rate of mistakes and hallucinations and it improves the quality of translations by having two AI engines discuss together until agreeing on a common translation.

On your side, I guess you used a single round with ChatGPT-4o because of the sheer volumes of entries, right?

sandakersmann:
Yes. I used ChatGPT-4o and asked for this:
`
Translate this to Thai. Keep same format and present as code: `

PR #1418, opened 2024-09-16
Add Option to Enable/Disable Screen Capture

denizt:
Veracrypt currently appears in screenshots and screen captures, which can unintentionally expose sensitive information, such as the fact that Veracrypt is running or the location of your volumes.

Both Windows and macOS offer mechanisms to exclude specific windows from being captured. While not foolproof, this is a useful preventative measure. The method is a no-op for Linux/FreeBSD.

For more details on the wxWidgets API, see: [wxTopLevelWindow::SetContentProtection](https://docs.wxwidgets.org/3.2/classwx_top_level_window.html#a337b9cec62b0cbd3b1b1545a83270f64)

### DIFF
--- a/src/Main/CommandLineInterface.cpp
+++ b/src/Main/CommandLineInterface.cpp
@@ -29,6 +29,7 @@ namespace VeraCrypt
 		ArgPim (-1),
 		ArgSize (0),
 		ArgVolumeType (VolumeType::Unknown),
+		ArgAllowScreencapture (false),
 		ArgDisableFileSizeCheck (false),
 		ArgUseLegacyPassword (false),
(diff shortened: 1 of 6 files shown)


idrassi:
Thank you for this interesting contribution. It definitely makes sense to protect against screenshots by default. 

I will merge this PR, although I plan to keep the `allow-screencapture` switch support only for macOS and Windows to avoid confusion for Linux users.

By the way, VeraCrypt for Windows doesn't use wxWidgets, but it would be useful to have this feature in Windows. I will check the source code of wxWidgets for inspiration on how to implement it.

PR #1451, opened 2024-12-03
Add PIV smart card keyfile encryption

the-dan:
Add an option to create a keyfile which is encrypted using a smart card's private key (e.g. YubiKey)


src/Common/Language.xml
```diff
@@ -203,6 +203,10 @@
     <entry lang="en" key="IDM_ADD_VOLUME_TO_SYSTEM_FAVORITES">Add Mounted Volume to System Favorites...</entry>
     <entry lang="en" key="IDM_ANALYZE_SYSTEM_CRASH">Analyze a System Crash...</entry>
     <entry lang="en" key="IDM_BACKUP_VOL_HEADER">Backup Volume Header...</entry>
+    <entry lang="en" key="IDM_REVEAL_REDKEY">Reveal Red Key...</entry>
+    <entry lang="en" key="REVEAL_REDKEY_INFO">This will produce decrypted keyfile which can be used without using the smart card, thus reducing security to two factors. This keyfile should be stored securely. Are you sure to continue?</entry>
+    <entry lang="en" key="REVEAL_REDKEY_DONE">Red key is saved</entry>
+    <entry lang="en" key="REVEAL_REDKEY_PATH">Choose file to save red key to</entry>
```
idrassi, line 209: Addition of new entries must always be done at the bottom of the XML file to preserve position of existing entries.

src/Common/Language.xml
```diff
@@ -367,6 +371,12 @@
     <entry lang="en" key="IDT_RANDOM_POOL_ENRICHMENT_NOTE">IMPORTANT: Move your mouse as randomly as possible within this window. The longer you move it, the better. This significantly increases security. When done, click 'Continue'.</entry>
     <entry lang="en" key="IDT_SECONDARY_KEY">Secondary key (hexadecimal)</entry>
     <entry lang="en" key="IDT_SECURITY_TOKEN">Security token:</entry>
+    <entry lang="en" key="IDC_SECURITY_TOKEN_KEY">Security token key</entry>
+    <entry lang="en" key="IDT_SECURITY_TOKENS">Security Tokens...</entry>
+    <entry lang="en" key="TOKEN_SLOT_ID">Slot ID</entry>
+    <entry lang="en" key="SELECT_TOKEN_KEYS">Select token keys</entry>
+    <entry lang="en" key="TOKEN_NAME">Token name</entry>
+    <entry lang="en" key="TOKEN_KEY_LABEL">Key label</entry>
```
idrassi, line 379: Same as above: new entries should be added at the bottom.

idrassi (changes requested):

src/Common/SecurityToken.cpp
```diff
@@ -188,7 +192,212 @@ namespace VeraCrypt
...
+				if (privateAttrib.size() == sizeof (CK_BBOOL) && *(CK_BBOOL *) &privateAttrib.front() != CK_TRUE)
+					continue;
+
+				GetObjectAttribute (slotId, dataHandle, CKA_MODULUS_BITS, privateAttrib);
+				if (privateAttrib.size() == sizeof (CK_ULONG)) {
+					CK_ULONG modulus = *(CK_ULONG *) &privateAttrib.front();
+					key.maxDecryptBufferSize = modulus / 8 - 11;
+					key.maxEncryptBufferSize = modulus / 8;
+				}
+
```
idrassi, line 275: This logic will fail for cards that contain ECC keys. Since the current implementation supports only RSA keys, the code should use CKA_KEY_TYPE to restrict logic to RSA keys only (CKK_RSA)

src/Common/SecurityToken.cpp
```diff
@@ -188,7 +192,212 @@ namespace VeraCrypt
...
+					continue;
+				
+
+				GetObjectAttribute (slotId, dataHandle, CKA_MODULUS_BITS, publicAttrib);
+				if (publicAttrib.size() == sizeof (CK_ULONG)) {
+					CK_ULONG modulus = *(CK_ULONG *) &publicAttrib.front();
+					key.maxDecryptBufferSize = modulus / 8 - 11;
+					key.maxEncryptBufferSize = modulus / 8;
+				}
+
```
idrassi, line 364: same as above: the code should restrict logic to RSA keys only.

src/Common/SecurityToken.cpp
```diff
@@ -361,7 +576,115 @@ namespace VeraCrypt
...
+	void SecurityTokenImpl::GetEncryptedData (CK_SLOT_ID slotId, CK_OBJECT_HANDLE tokenObject, vector <uint8> plaintext, vector <uint8> &ciphertext)
+	{
+		LoginUserIfRequired (slotId);
+
+		if (Sessions.find (slotId) == Sessions.end())
+			throw ParameterIncorrect (SRC_POS);
+
+		CK_MECHANISM m = { CKM_RSA_PKCS, 0, 0 };
+		CK_RV status = Pkcs11Functions->C_EncryptInit (Sessions[slotId].Handle, &m, tokenObject);
```
idrassi, line 644: We should not use CKM_RSA_PKCS since it is insecure. We must use CKM_RSA_PKCS_OAEP with SHA256 MGF.

src/Common/SecurityToken.cpp
```diff
@@ -361,7 +576,115 @@ namespace VeraCrypt
...
+
+	void SecurityTokenImpl::GetDecryptedData (CK_SLOT_ID slotId, CK_OBJECT_HANDLE tokenObject, vector <uint8> edata, vector <uint8> &decryptedData)
+	{
+		LoginUserIfRequired (slotId);
+
+		if (Sessions.find (slotId) == Sessions.end())
+			throw ParameterIncorrect (SRC_POS);
+
+		CK_MECHANISM m = { CKM_RSA_PKCS, 0, 0 };
+		CK_RV status = Pkcs11Functions->C_DecryptInit (Sessions[slotId].Handle, &m, tokenObject);
```
idrassi, line 672: Like above, we should not use CKM_RSA_PKCS since it is insecure. We must use CKM_RSA_PKCS_OAEP with SHA256 MGF.

idrassi (changes requested):

idrassi:
@the-dan Thank you for this contribution.
Native support for PKI smart cards has always been planned but I personally never found the time to work on it. Your approach strikes a good balance between native support and the keyfiles mechanism, making it a strong first step.

I will need some time to review the code changes because there are so many...this is the largest PR I have ever had to review.

That being said, I have already provided some comments and will add more as I progress with the review.

the-dan:
@idrassi I’m happy that this change aligns with VeraCrypt’s direction.

Thanks for going through this! I haven’t touched the encryption part since the initial prototype started working, which is why plain RSA PKCS was used.

Instead of selecting only RSA keys and locking in the mechanism, I revisited the approach to make it more flexible. For now, it supports only RSA OAEP, but I believe it could be extended in the future.

Please take a look when you have time. Apologies for the large commits—this time, I also did a lot of renaming for clarity, which affected many files

war59312:
DOA?

idrassi:
@war59312 This feature is not dead. It is just the amount of work necessary for its integration is not trivial and there are issues that have more demand from user community. Of course, if someone is willing to sponsor work on it, it can get more priority.
Anyway, I will try to revive this PR after I finish preparing 1.26.28 version.

PR #1471, opened 2025-01-23
Update Language.pl.xml

p0k33m0n:
Improved formatting.


idrassi:
Other language files use spaces on those two entries and they don't cause any issue.

Did you notice any problem with the spaces?

I think they don't change anything on the UI.

p0k33m0n:
It expands the menu in such a way that it does not display everything. In fact, the entire upper beam is falling apart. I don't know where the idea to add spaces came from.

idrassi:
Interesting behavior. Maybe it has to do with DPI scaling.

I don't recall the history behind adding spaces...

Can you share a screenshot so that I have an idea about the issue?

p0k33m0n:
obraz

Original screenshot, without modifications. Removing the spaces fixes the problem.

idrassi:
I see. The font for the menu bar is obviously too large compared to the static size of VeraCrypt dialog.
This the main problem of VeraCrypt UI: it is static, not dynamically adapting to text size.
Below in how it looks on my machine.

Anyway, I will merge your PR. This is the best solution for now.

image

PR #1478, opened 2025-01-29
refactor: use the term unmount instead of dismount

tessus:
This PR has more than 1 commit on purpose and can be squashed during merge. (Or I can squash them as well, even though I like the commits separate in PRs (and not force push in a PR) to make discussion/comments easier to follow. But I will adhere to whatever process this project follows.)

1. use `UNMOUNT` in constants
2. use the term unmount in the GUI
3. update the documentation
4. add another command line argument/flag `unmount` (`dismount` will still work)

This is a rather extensive change in terms of LOC changed. I split it in several commits to make the review a bit easier.

I haven't changed variable names for now, but can do so upon request. I have noticed that in certain parts of the code the term `unmount` had already been used.


doc/html/Command Line Usage.html
```diff
@@ -74,20 +74,20 @@ <h1>Command Line Usage</h1>
...
 <tr>
 <td><em>/auto</em> or <em>/a</em></td>
 <td>If no parameter is specified, automatically mount the volume. If devices is specified as the parameter (e.g., /a devices), auto-mount all currently accessible device/partition-hosted VeraCrypt volumes. If favorites is specified as the parameter, auto-mount
  favorite volumes. Note that /auto is implicit if /quit and /volume are specified. If you need to prevent the application window from appearing, use /quit.</td>
 </tr>
 <tr>
-<td><em>/dismount</em> or <em>/d</em></td>
-<td>Dismount volume specified by drive letter (e.g., /d x). When no drive letter is specified, dismounts all currently mounted VeraCrypt volumes.</td>
+<td><em>/unmount</em> or <em>/d</em></td>
+<td>Unmount volume specified by drive letter (e.g., /d x). When no drive letter is specified, unmounts all currently mounted VeraCrypt volumes.</td>
```
idrassi, line 86: The documentation entry should be corrected: /unmount comes with shortened version /u and deprecated /dismount is still supported with shortened /d.
I proposed to keep an entry for /dismount that indicate that it is deprecated and similar to /unmount and then add new entry for /unmount with correct shortened version.
For example:

```
<tr>
<td><em>/dismount</em> or <em>/d</em></td>
<td>Deprecated. Similar to /unmount or /u.</td>
</tr>
<tr>
<td><em>/unmount</em> or <em>/u</em></td>
<td>Unmount volume specified by drive letter (e.g., /u x). When no drive letter is specified, unmounts all currently mounted VeraCrypt volumes.</td>
</tr>
```

idrassi (changes requested):

doc/html/Command Line Usage.html
tessus: Thanks for reminding me. It was on my list, but I forgot to do that. Will get to it soon.

tessus: Done.

idrassi (approved):

idrassi:
@tessus Thank you!

tessus:
You are welcome and thank you for accepting this change.

idrassi:
By "IT standards," I meant what is commonly used as a standard term for disconnecting a disk or volume, not an official standard like ISO or similar.

The term "dismount" originated in the Windows ecosystem where it was chosen in contrast to "unmount" the term used in UNIX and macOS environments.
However, Microsoft has since transitioned to using "unmount" as the official term, aligning with all other operating systems. Some internal APIs and constants still retain "dismount" for backward compatibility but the shift toward "unmount" is now standard practice.

For further clarification, this article explains the distinction between the two terms: https://thecontentauthority.com/blog/unmount-vs-dismount

Additionally, the Cambridge Dictionary definitions reinforce the difference:

- https://dictionary.cambridge.org/dictionary/english/unmount
- https://dictionary.cambridge.org/dictionary/english/dismount

It was time for VeraCrypt to move away from "dismount" and I thank @tessus for helping make this change happen.

apawelec:
Maybe I'm missing something, but most obvious un/dis-mount button still has old label:

obraz

Installed from veracrypt-1.26.20-Ubuntu-22.04-amd64.deb.

tessus:
I changed it though:

https://github.com/veracrypt/VeraCrypt/pull/1478/commits/700a24a78554b43bd56b1a6587157f68256c43f1#diff-3df32839fe2f3289e5aad00e28aa5a7f6cac1c4ff6d2880395be314f7c7e9bffR1012

image

tessus:
I've checked the entire code again (current master). There is no string of `Dismount All` in the source tree.

But @apawelec is correct. This string also shows in the macOS version.

@idrassi how are the packages built? Is it possible that there was an issue with the build process?

idrassi:
@apawelec Thank you for reporting this. It is amazing how we all missed this button...the brain can sometimes play tricks on you!

@tessus 
The issue is caused by the entry IDC_UNMOUNTALL in Language.xml that used &amp; inside it to indicate that `ALT+s` is accelerator key associated with unmounting all volumes:

``` xml
    <entry lang="en" key="IDC_UNMOUNTALL">Di&amp;smount All</entry>
```

Fixing this will require using "Unmount All" and since it doesn't contain an '`s`', the usual accelerator key cannot be used anymore. The logic approach is to replace it with `ALT+u` since it is more logical and it is not currently used. Honestly, `ALT+s` never made sense since `s` is more associated with saving.

I will push a change to fix this and use `ALT+u` as a new accelerator key for unmounting all volumes.

idrassi:
Fix pushed: https://github.com/veracrypt/VeraCrypt/commit/42b716f8472ace23b13e609fe920267cecebac84

PR #1479, opened 2025-01-30
Enable AESNI only on x86

Xeonacid:
This helps building on riscv which does not have immintrin.h.


idrassi:

Issue #1487, opened 2025-02-07
Project "Crypto" only support x64, but some Win32 Project need Crypto.lib


Nebudust:
Hello, I'm using Visual Studio 2022 to compile VeraCrypt.
There are many compiling errors, but I fixed them, but I can't find where "ALL Debug/Release" is.
When I select "x64" and try to build solution, I got 3 exe files. Then I try to select "Win32" and build again, project "COMReg" said it need ..\Crypto\Release\Crypto.lib. I found it at ..\Crypto\x64\Release but it is a x64 lib. I checked the target and then found project "Crypto" does not support "Win32" and "COMReg" has only "Win32". I don't know how to continue build...

![Image](https://github.com/user-attachments/assets/fe9ca0f1-67a5-4d4b-a979-6107d207805b)

![Image](https://github.com/user-attachments/assets/af2c5bc9-bd78-4cd1-b5b1-495e7ba827dd)

VeraCrypt Source : 1.26.20
Visual Studio Version: VS2022
OS: Windows 11
Environment: VMWare

idrassi:
Thank you for reporting this.

Indeed, there is an issue in the "Crypto" project: its "Win32" configuration was removed and so the function sha512 is not available anymore and the depending projects fails because crypto.lib cannot be found.

The easy fix is to add back the Win32 configuration to Crypto project. But the best approach (and this is what I had in mind when I removed Win32 configuration) is to implement sha512 in a standalone way for the 32-bit build so that we don't need Crypto project.

I have pushed a commit that implements this approach: https://github.com/veracrypt/VeraCrypt/commit/a2f11e11024a104dabf6921cd276baa4255473e0

Please use latest code from master branch and try again.

Nebudust:
I opened newest master branch in VS2022, and build failed.

Error from here
ComSetup.cpp:
#include "../Mount/MainCom_i.c"
#include "../Format/FormatCom_i.c"

By Add "Mount" project reference to "SetupDLL", all build is done
I think this is the last problem for VS2022 with 1.26.20

Nebudust:
I had open a PR (https://github.com/veracrypt/VeraCrypt/pull/1490) to fix this problem, please check it

idrassi:
Indeed, these files require the type lib files to be present but only to get the GUIDs.
[...]

PR #1490, opened 2025-02-08
Fix SetupDLL Include Error

Nebudust:
Fix SetupDLL Include Error:
ComSetup.cpp:
#include "../Mount/MainCom_i.c"
#include "../Format/FormatCom_i.c"


idrassi:
Thank you.
As indicated in the issue https://github.com/veracrypt/VeraCrypt/issues/1487, I have pushed a different fix that avoids adding extra dependencies during build: https://github.com/veracrypt/VeraCrypt/commit/974a3909ecaf0de63f7ef36b2ed6d1e0a7804267

PR #1495, opened 2025-02-11
Fix SetupDLL and ExpandVolume Error

Nebudust:
1. Add SETUP_DLL to ReleaseCustomEFI
2. Change link output file name to "VeraCryptSetup.dll"
3. Fix ExpandVolume's lib profile error ("AMD64" -> "ARM64")


idrassi:
@Nebudust Thank you for the efforts and the fix.

PR #1497, opened 2025-02-15
CI: Enable XML Validation Workflow

Jertzukka:
Enable workflow on correct branch and make script executable.


idrassi:
Welcome back! 
Thank you for the fix and of course for the code since it is yours.

harryytm:
import pkg from '../../package.json' assert { type: 'json' };
                                     ^^^^^^

SyntaxError: Unexpected identifier 'assert'
    at compileSourceTextModule (node:internal/modules/esm/utils:338:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:106:18)
    at #translate (node:internal/modules/esm/loader:473:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:520:27)
    at async ModuleJob._link (node:internal/modules/esm/module_job:115:19)

Node.js v23.8.0

idrassi:
@harryytm Thanks for the report. It looks like the new Node.js version 23.8 is causing an issue with fast-xml-parser. Since it was working before, I think the solution is to force the use of LTS version of Node.js (20.x or 22.x for example).

I will try first to reproduce the issue on my side before pushing any changes.

idrassi:
I have done tested under Ubuntu 24.04 with latest Node.js 23.8 and I discovered that the issue is caused by the latest version series 5.x of fast-xml-parser: if I use version 4.5.2 which is the latest of 4.x series, the issue doesn't occur.

So I will push a change to force the use of version 4.5.2 of fast-xml-parser in this Github workflow.

PR #1509, opened 2025-03-03
Fix windows target and atls.lib error

Feryquence:
I don't know if it is correct, but without its compiler returns "No target architecture"
and fixed atls.lib error in custom EFI


idrassi:
I don’t encounter most of the issues you mentioned. The only valid issue I found is the missing `atls.lib` dependency in the Portable, Setup, and SetupDLL projects. I have no issues with the driver build, nor do I need to add `atlsd.lib`, so it seems there may be a problem with your build environment.

I can’t merge this PR in its current state due to these discrepancies. If you update it to include only the addition of the missing `atls.lib`, I’ll be happy to merge it.

idrassi:
Unfortunately, I didn't receive an answer about my query to update this PR. 
So I have pushed a commit that includes the addition of atls.lib to Portable, Setup, and SetupDLL projects in custom EFI configuration: https://github.com/veracrypt/VeraCrypt/commit/08093f460ad210f8e5f38cd0576f72cbc59dee65

Closing this PR.

PR #1514, opened 2025-03-12
Linux/FreeBSD: Improvements on installation scripts

Jertzukka:
Few improvements on Linux/FreeBSD installation and package scripts, and user experience when using generic installer:
- Use correct files in FreeBSD installation scripts and separate the files from the Linux counterparts to make differing changes clear
- Ensures mime-types and desktop files are updated on the system when generic installer is used as we're not relying on a package distributor to do it
- Ensure the files are properly removed through `veracrypt-uninstall.sh` before printing success
- Move from deprecated `-e` syntax to `--` in gnome-terminal
- Adds title to xmessage windows
- Remove unsupported konsole `--caption` option from FreeBSD installation script

### DIFF
--- a/src/Main/Main.make
+++ b/src/Main/Main.make
@@ -179,7 +179,7 @@ endif
 GUI_CONDITION := $(filter gui,$(INSTALLER_TYPE))
 GTK2_CONDITION := $(filter 2,$(GTK_VERSION))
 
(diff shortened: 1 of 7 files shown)


idrassi:
Thank you for the fixes and enhancements. 
Very good PR as always 👍

PR #1516, opened 2025-03-22
fix Language.hu.xml invalid character

balintx:
This patch fixes the invalid XML in the Hungarian translation XML file, which cause the following error when trying to launch Veracrypt:

ParameterIncorrect at VeraCrypt::XmlParser::GetNodes:82

### DIFF
--- a/Translations/Language.hu.xml
+++ b/Translations/Language.hu.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <VeraCrypt>
   <localization prog-version= "1.26.20">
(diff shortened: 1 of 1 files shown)


idrassi:
@balintx  Thank you for reporting this issue. Actually the Hungarian XML file content is valid and the presence of '>' character is correct. The issue is on the XML parsing code in the file Main/Xml.cpp file: the method GetNodes implementation doesn't parse multiple attributes correctly.

I have implemented a fix for this method and I will be pushing it soon.

This means that this PR will not be merged since there is no need modifying the valid Hungarian XML file.

I will update here when I push the fix.

idrassi:
I have pushed a fix for the XML parser that was causing the issue: https://github.com/veracrypt/VeraCrypt/commit/56301ea64b8bce666ab1fd51f1d311c549af1749
I will close this PR.
Thank you again for the report.

PR #1523, opened 2025-04-12
User Guide with THML

fzxx:
Hi @idrassi , I have finished translating the user guide and hope to see it in https://veracrypt.fr/zh-cn/


idrassi:
Thank you for the amazing work!

idrassi:
@fzxx I have uploaded the translated documentation to the website and I also translated other pages in the website using AI in order to have better browsing.

Please check it https://veracrypt.fr/zh-cn/.

Let me know if you find any mistakes in the translated website pages

fzxx:
> @fzxx I have uploaded the translated documentation to the website and I also translated other pages in the website using AI in order to have better browsing.
> 
> Please check it https://veracrypt.fr/zh-cn/.
> 
> Let me know if you find any mistakes in the translated website pages

When accessing https://www.veracrypt.fr and redirecting to https://veracrypt.fr/en/Home.html, there is no redirection carried out based on the recognition of the browser language or the system language.

idrassi:
Indeed, there is no automatic redirection. We cannot simply detect browser language: we will need to add a way for users to choose correct language in case of wrong detection or if there are multiple languages in the browser. This will need to add javascript code and modify all pages to add language selection.

This is not simple work, and it needs time.

fzxx:
> Indeed, there is no automatic redirection. We cannot simply detect browser language: we will need to add a way for users to choose correct language in case of wrong detection or if there are multiple languages in the browser. This will need to add javascript code and modify all pages to add language selection.
> 
> This is not simple work, and it needs time.

It may be easy to do this with AI

fzxx:
> I have implemented a simple solution where language selection override is only in Home.html. This should be enough since automatically redirection from https://veracrypt.fr redirects to Home.html. For other URLs, it is the user who selected them so no need to override.
> 
> Please open https://veracrypt.fr and share your feedback. Thank you.

Very good, no problem with the basic functions;
It is recommended that the VeraCrypt User Guide_zh-cn.chm file be added to the installation package for the next update.

idrassi:
Thanks for the confirmation.

I have already pushed changes to include the CHM in the installation although I renamed it to VeraCrypt User Guide.zh-cn.chm with a dot to be compatible with the code that looks for localized CHM files.

I also changes the organization of folders, and I made several fixes.

You can get the latest code from master to see what changed.

fzxx:
> Thanks for the confirmation.
> 
> I have already pushed changes to include the CHM in the installation although I renamed it to VeraCrypt User Guide.zh-cn.chm with a dot to be compatible with the code that looks for localized CHM files.
> 
> I also changes the organization of folders, and I made several fixes.
> 
> You can get the latest code from master to see what changed.

I noticed
https://github.com/veracrypt/VeraCrypt/tree/master/doc/chm/zh-cn file has not changed, it is recommended to change it to VeraCrypt.zh-cn

idrassi:
Done in https://github.com/veracrypt/VeraCrypt/commit/e946de72413b9d0927a053f92ec17396790e6129

Thanks

Issue #1521, opened 2025-04-06
change the default mount point under `/media` or `/run/media` instead of `/mnt`


evan0greenup:
# Current behavior

When mounting a file or a device, `veracrypt` will create a directory under `/mnt`.

# The drawback of current behavior

The directory `/mnt` is usually a place where user want to manually operate on. For example, mounting a `chroot` linux distro root filesystem (Arch Linux), or mounting a specific filesystem which user want to operate it inside shell.

The nature of mounting which `veracrypt` need is same as what `udisk2` like (Just a place where user want to open inside file manager).

# Desired behavior

As a result, it would be nice to follow the same or similar behavior as udisk2 or other auto-mounting facility of Linux system. Creating a mountpoint under `/media` or `/run/media` and under the subdirectory of username. This will make `/mnt` available during mounting of `veracrypt` volume.

# Screenshots/Mockup/Designs

# Additional information

# Your Environment

Please tell us more about your environment

**VeraCrypt version:** 

**Operating system and version:** 

**System type:**

Jertzukka:
Currently `/media/` is used preferentially if it exists, after that it fallbacks to `/mnt/` and then temporary files.

https://github.com/veracrypt/VeraCrypt/blob/2f8161af342217a5f59c4b7d6c73f425857a618b/src/Core/Unix/CoreUnix.cpp#L288-L301

evan0greenup:
@Jertzukka , I think it would be good to add `/run/media` and remove `/mnt`.

Because for nowadays Linux distro, almost certain either `/media` or `/run/media` will do exist. 

And if mounting on a temporary directory, mount-point actually doesn't take up storage space and won't consume tmpfs.

But if a user doing chroot inside `/mnt`, mounting in `/mnt/veracrypt` will bring a lot of undetermined inconvenient to the user.

stevenleeS0ht:
@evan0greenup , will <https://github.com/veracrypt/VeraCrypt/pull/1524> help?

stevenleeS0ht:
To be note that the long or short of prefix does not matter at all, because, the most common way that user enter the directory is by clicking the drive row entry in VeraCrypt GUI.

evan0greenup:
> [@evan0greenup](https://github.com/evan0greenup) , will [#1524](https://github.com/veracrypt/VeraCrypt/pull/1524) help?

Yeah, I like it!!!

PR #1524, opened 2025-04-13
modify unix default mounting point prefix (use /run/media/veracrypt t…

stevenleeS0ht:
…o replace /mnt/veracrypt).

This pull request is to solve this issue <https://github.com/veracrypt/VeraCrypt/issues/1521>.


stevenleeS0ht:
@idrassi , please have a look

src/Core/Unix/CoreUnix.cpp
```diff
@@ -294,8 +294,8 @@ namespace VeraCrypt
 		if (FilesystemPath ("/media").IsDirectory())
 			return "/media/veracrypt";
 
-		if (FilesystemPath ("/mnt").IsDirectory())
-			return "/mnt/veracrypt";
+		if (FilesystemPath ("/run/media").IsDirectory())
+			return "/run/media/veracrypt"
```
Jertzukka, line 298: This won't compile without the semicolon:
```suggestion
			return "/run/media/veracrypt";
```

While I think /run/media is the right modern location for automatically mounted devices, I think /mnt should still be kept as a fallback. The reason is that placing the mountpoint into /tmp should be avoided, and there are still some old systems (such as Slackware) which veracrypt is ported over to which don't have /media nor /run in the file hierarchy as they still use /var for it.

stevenleeS0ht:
@Jertzukka , I fix the semicolon and append `\mnt` to the last one before temporary directory. Is it OK now?

stevenleeS0ht:
@idrassi  is it ok?

idrassi:
@stevenleeS0ht Thank you for the proposed changed.

Using /run/media on Linux is ok for me since it is closer to the approach adopted by udisk2. I would ask to put the new code inside `#if defined (TC_LINUX)` `#endif` since this is Linux specific.

Can you please do this change?

Another thing is to check the behavior of FileManagers like Nautilus, Nemo and Dolphin after applying this change. For example, today on Linux Mint, after mounting a volume, it is automatically detected and shown in Nemo FileManager. I will try to check the behavior after this change.

stevenleeS0ht:
> @stevenleeS0ht Thank you for the proposed changed.
> 
> Using /run/media on Linux is ok for me since it is closer to the approach adopted by udisk2. I would ask to put the new code inside `#if defined (TC_LINUX)` `#endif` since this is Linux specific.
> 
> Can you please do this change?
> 
> Another thing is to check the behavior of FileManagers like Nautilus, Nemo and Dolphin after applying this change. For example, today on Linux Mint, after mounting a volume, it is automatically detected and shown in Nemo FileManager. I will try to check the behavior after this change.

@idrassi , I think `#ifdef TC_LINUX` like other places, would be more concise.

idrassi:
Thank you for the change.

I'm having a hard time testing this modification because I haven’t found a single Linux distribution where it gets triggered. Either /media already exists, so /run/media is not checked, or neither /media nor /run/media exist, in which case /mnt is used.

In what environment would this change be useful? Given that we already have the `VERACRYPT_MOUNT_PREFIX` environment variable to override the default mount point prefix, we need a strong justification for this addition.

evan0greenup:
@idrassi @stevenleeS0ht , Sorry for forgetting put Linux Distro information in my issue.

I can confirm that Arch Linux and all its derivatives use `/run/media` as prefix. This is quite a large portion of Linux world.

evan0greenup:
@idrassi , as evidence, please read <https://wiki.archlinux.org/title/Udisks#Mount_to_/media>

> By default, udisks2 mounts removable drives under the ACL controlled directory /run/media/$USER/

stevenleeS0ht:
> Thank you for the change.
> 
> I'm having a hard time testing this modification because I haven’t found a single Linux distribution where it gets triggered. Either /media already exists, so /run/media is not checked, or neither /media nor /run/media exist, in which case /mnt is used.
> 
> In what environment would this change be useful? Given that we already have the `VERACRYPT_MOUNT_PREFIX` environment variable to override the default mount point prefix, we need a strong justification for this addition.



@idrassi , yeah, I also use Arch Linux (EndeavourOS for desktop), the information and links provided by @evan0greenup is correct.

idrassi:
Thank you for these details.

I also did a check on an Arch Linux machine and /media is absent which makes VeraCrypt use /mnt. With this change, /run/media will be used which will align with expected logic.
I don't see any issues with this change so I'm merging it.

PR #1526, opened 2025-04-15
Add alignment constants for derived key and KEY_INFO buffers to ensure SIMD compatibility.

bernardladenthin:


idrassi:
Thank you for this additional change.

PR #1529, opened 2025-04-21
Fix the whirlpool for big-endian platform

pingwcy:
Fix the endian issue for Whirlpool Hash on big-endian platform


idrassi:
Thank you for the fix.

PR #1530, opened 2025-04-23
Fix a mistake

fzxx:


idrassi:
Thanks for the update. Indeed, the content was from another page.

PR #1547, opened 2025-05-29
Documentation: Remove XHTML spec and fix errors

Jertzukka:
Both the local documentation and the one on the website advertises itself as adhering to XHTML 1.0 Transitional specification, which is not the case. This becomes an issue when programs, such as the default web browser on GNOME, Gnome Web expects the page to follow the set spec. This causes nearly all pages of the documentation to fail to render at least partly.

You can also see this through online validator such as https://validator.w3.org/#validate_by_uri and giving it any documentation page like https://veracrypt.io/en/Documentation.html.

PR also fixes some issues I found while fixing the HTML. CHM guides were also generated to be up-to-date.


idrassi:
Thank you @Jertzukka for fixing this. Can you please remove the CHM files from your PR? This will solve the conflict and also I cannot accept CHM files from external sources since they are compiled code.
Unfortunately, I could not take your PR into account in time for the 1.26.24 release.

Jertzukka:
@idrassi Should be good now.

idrassi:
Thank you very much! I appreciate your help and contributions.

PR #1568, opened 2025-07-03
Fix IDC_MOUNTALL in German translation

helmutbuhler:
In German, the UI looks like this:
grafik
The second button means: "Mount all devices...". That's inconsistent with the english wording "Auto-Mount Devices".
A correct literal translation would be "Datenträger automatisch einhängen...", but that's too long, even if abbreviated. So I change this in this PR to this:
grafik
which means something like "Automatically mount..."

Furthermore, this PR removes the space before all the "..." in strings. That makes it consistent with the english version, and all other german texts. (This is in a separate commit)


idrassi:
Thank you for this enhancement.

PR #1569, opened 2025-07-05
Initial work on Ocrypt integration

waywardgeek:
I'll need feedback on this PR to get it to the point that it is acceptable to the VeraCrypt authors.  This version works in --text mode on Linux when creating and mounting volumes encrypted via Ocrypt. OpenSSL dependency.  VeraCrtypt is missing bignum, AES-GCM, HKDF, and elliptic curve support for ed25519.  How should I resolve this?  E.g. is libnacl acceptably lightweight?

Ocrypt requires network access, making non-compatible with full-disk encryption, and it also has clear metadata I currently store in the header after address 512, defeating plausible deniability.  Ocrypt must remember the URLs used to register its secret, so I see no way around this limitation.

The code is currently sloppy.  However, it works and is a nice PoC.  How should we proceed?


Neustradamus:
@idrassi; Have you seen this @waywardgeek PR?

idrassi:
@waywardgeek

Thank you for putting this together and sorry for the delayed feedback. The project had been stalled because of various issues and I am only now resuming activities.

After reviewing the PR and Ocrypt/OpenADP, I have several reservations about integrating this approach into VeraCrypt core. As a result, I do not think this PoC can be merged into VeraCrypt in its current form.

As you noted, several aspects depart from VeraCrypt's philosophy and threat model, in particular the loss of plausible deniability and the addition of network access. After discussions with others, I do not think integrating such a feature directly into VeraCrypt core would be acceptable, even if it were not enabled by default.

Separately, I am considering a narrower extension mechanism for VeraCrypt: user-selected virtual keyfile providers, similar in spirit to the existing EMV/security-token keyfile support.

In that model, VeraCrypt would keep the standard volume format, header KDFs, XTS mode, master-key handling and header validation unchanged. A provider would not be auto-detected. The user would explicitly select it, VeraCrypt would render a limited declarative input form supplied by the provider and the provider would return virtual keyfile bytes that VeraCrypt mixes through the existing keyfile path.

This could be a cleaner place for mechanisms such as OpenADP. OpenADP could be installed as a third-party extension by interested users and would act as an explicitly selected online virtual-keyfile provider rather than as a VeraCrypt PRF or a new header format. Any required OpenADP metadata would remain outside the VeraCrypt volume unless the user knowingly accepts the loss of plausible deniability.

This is a separate design direction and not something this PR can resolve, but feedback on such a virtual-keyfile-provider model would be welcome.

I will keep this PR open for now for discussion, although I do not intend to merge the current PoC.

PR #1590, opened 2025-08-29
Fix problems inside crash dump filter path

Chuccle:
* Prefer allocations to be non-executable
* Remove and reimplement DDIs inappropriately called inside HIGH_LEVEL IRQL routines
* Refactor hibernate context to be passed around in the passed FILTER_EXTENSION pointer rather than global

### DIFF
--- a/src/Driver/DumpFilter.c
+++ b/src/Driver/DumpFilter.c
@@ -16,16 +16,41 @@
 #include "Tests.h"
 #include "cpu.h"
 
-static DriveFilterExtension *BootDriveFilterExtension = NULL;
-static LARGE_INTEGER DumpPartitionOffset;
-static uint8 *WriteFilterBuffer = NULL;
-static SIZE_T WriteFilterBufferSize;
+typedef struct _DumpFilterContext
+{
+	DriveFilterExtension* BootDriveFilterExtension;
+	LARGE_INTEGER DumpPartitionOffset;
+	uint8* WriteFilterBuffer;
+	SIZE_T WriteFilterBufferSize;
+	PMDL WriteFilterBufferMdl;
+} DumpFilterContext;
 
+static void Cleanup(DumpFilterContext* dumpContext)
+{
+	if (!dumpContext)
+		return;
+
+	if (dumpContext->WriteFilterBufferMdl)
+	{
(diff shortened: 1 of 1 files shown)


Chuccle:
Open to clarifying any proposed changes if needed.

idrassi:
Thank you for these proposed enhancements.

While the idea is good, the current PR contains some issues:

- The code doesn't build: `Cleanup` function is used before its declaration so there is a warning that causes build to fail. At least a forward declaration is needed before using it.
- The PFN array memcpy should be preceded by an explicit boundary check. The comment says "dstPfnCount should always be >= srcPfnCount" but I cannot seem to derive this guarantee when reading the code (we only have dataLength <= WriteFilterBufferSize). can you please clarify?
- The `WriteFilterBufferMdl` is allocated but never built: there is a missing call to `MmBuildMdlForNonPagedPool`.
- The PR modifies critical MDL fields (`MappedSystemVa`, `StartVa` and `ByteOffset`). This looks dangerous to me. Do we really need to change these? The copy seems enough.
- The PR uses an `ASSERT `for MDL mapping state in `DumpFilterWrite` but in Release build the `ASSERT` is removed. It makes more sense to have a real runtime check.

Looking forward for your feedback.

Chuccle:
Hi, 

Yep some sloppiness on my part regarding MDL building and ensuring it compiles.

For the PFN array copying, the filterInitData->MaxPagesPerWrite should dictate the maximum pages represented by a given "writeMdl". 

However, I "could" technically do some defensive programming `memcpy(dstPfnArray, srcPfnArray, min(dstPfnCount, srcPfnCount) * sizeof(PFN_NUMBER));` However, this would obviously fail hibernate anyway but would maybe avoid a access violation event in event viewer I suppose.

The assertion is there as a sanity check. In HIGH_LEVEL IRQL, we absolutely must have memory be resident and mapped.

Modifying the MDL structure "header" is admittedly not ideal, what i've shown here is a functional equivalent to the MmInitializeMdl DDI, however, you'll notice in the documentation that it is intended for <= DISPATCH_LEVEL (although obviously it works at any level as currently implemented), what's to stop MS from changing it in future to make it break in crash/hibernate context?

Issue #1572, opened 2025-07-15
Issues with Windows USB energy saving (it ignores/fails to detect VC volume in use)


JustMyGithub:
# Expected behavior
When a drive is used, USB energy is not cut off

# Observed behavior
After some time of continuous write, veracrypt-hosting usb hdd is disconnected (and therefore dismounted) and the copy fails.
It may only happen with locked OS. Sometimes it was after less than 20 Minutes, sometimes it took more than 40 minutes.

# Steps to reproduce

0. Have energy management for usb devices turned on in windows. Afair its default.
1. copy a large file from a bitlocker usb-hdd to a veracrypt usb-hdd
2. (maybe required: lock screen)
3. wait

# Your Environment
Windows 10, VC 1.26.24, 64 bit

I tested 3 usb drives, 2 cables and 2 usb ports (cables and ports all combinations). Bitlocker (source) drive was never affected, but veracrypt one was several times. After disabling usb device energy management in windows energy settings, it immediately worked for several hours without issues. 

So I assume that windows can detect the hdd corresponding to a bitlocker drive (and its activity and hence not cut usb power off) but not the hdd corresponding to a veracrypt drive and tehrefore it cuts of the usb port if energy saving for usb is enabled.

I am nut sure whether this can be fixed by VC or only on the windows side, but definitely it is a bug.

andry81:
That's a long living bug: https://github.com/veracrypt/VeraCrypt/issues/1443

JustMyGithub:
It is not the same issue:
- This is a partition, 1443 is a container. 
- This does dismount automatically, 1443 does not dismount, even if requested. 
- 1443 is a 3,5" usb drive case which means it is powered externaly, not via USB, like the extranl hdds I used.

andry81:
> BUT I suggest you try to disable energy saving for USB and test your AgeStar 3U2B3A and check whether this solves your issue.

These issues some how related to the USB connection, not just to a particular device. Because the issue is catched on other devices attached to the USB.

Issue #1443, opened 2024-11-17
Failure on dismount a container file on sleeping external USB drive


andry81:
1. External USB drive is connected through AgeStar 3U2B3A
2. Container file on the drive is attached
3. Drive is sleeping
4. Try to dismount the container

VeraCrypt time to time fails to dismount the container with an error dialog said something `Failure dismount` and immediately force dismounts the container (popup balloon message).

# Expected behavior

Keep waiting the drive spin up.

# Observed behavior

Force dismount on some event or timeout.

# Your Environment

Please tell us more about your environment

**VeraCrypt version:** 1.26.7 64-bit

**Operating system and version:** Windows 8.1 9600

**System type:** 64-bit

PR #1601, opened 2025-09-06
Fix problems with devices changing power states while mounted

Chuccle:
Fixes: #1572
Fixes: #1443


src/Driver/DriveFilter.c
```diff
@@ -987,12 +987,30 @@ static NTSTATUS DispatchPower (PDEVICE_OBJECT DeviceObject, PIRP Irp, DriveFilte
...
+		{
+			if (irpSp->Parameters.Power.Type == DevicePowerState)
+			{
+				DEVICE_POWER_STATE requestedState = irpSp->Parameters.Power.State.DeviceState;
+
+				// Is this too broad?
+				if (requestedState > PowerDeviceD0 && Extension->DriveMounted)
+				{
+					IoReleaseRemoveLock(&Extension->Queue.RemoveLock, Irp);
+					return TCCompleteIrp(Irp, STATUS_DEVICE_BUSY, 0);
```
Chuccle, line 1008: I'm going to swap this to an NT_ERROR range code - STATUS_DEVICE_BUSY, although appropriate, is NT_WARNING range. I don't think that's right to complete with.

PR #1626, opened 2025-11-15
Update Language.ro.xml

TigerxWood:
Update and proofreading Romanian


TigerxWood:
In line 973, RESCUE_DISK_CHECK_FAILED, there is a variable place holder ( '%s') after '...please try another medium%s.' Is this correct or a mistake?

PR #1643, opened 2026-01-05
Update missed links and period

OSINTELSOLUTIONS:


OSINTELSOLUTIONS:
Please change the project link on https://wikipedia.org/wiki/VeraCrypt to [veracrypt.jp](https://veracrypt.jp/) in all languages

idrassi:
We cannot change the content of past release notes: they are part of historical changes that are tied to specific versions. The presence of .fr link for old versions release notes is ok. 

As for Wikipedia, I just saw that someone already corrected the official link.

I'm closing this PR. Thank you for your feedback.

Issue #1426, opened 2024-10-06
Dismounting a drive from command line does not release the drive letter


DavidOfMA:
# Expected behavior

Dismounting the drive from the command line should work the same as from the UI and release the drive letter. This is how it worked in previous versions of Veracrypt and how it still works if I dismount from the UI.

# Observed behavior

When I dismount a drive using the command line (for example, "C:\Program Files\veracrypt\veracrypt.exe" /dismount J: /q), the drive letter is not released. Explorer sees it as a Local Disk, even after the drive is removed with Safely Remove. 

# Steps to reproduce

1. Mount an external disk drive in Veracrypt and assign it a drive letter J:
2. "C:\Program Files\veracrypt\veracrypt.exe" /dismount J:
3. The drive is dismounted but still appears as a Local Disk J: and cannot be re-assigned to J: in Veracrypt
4. Remove the drive with Safely Remove
5. The drive is still listed in Explorer as Local Disk J: but is invalid. 

# Screenshots

# Your Environment
Windows 11 Pro, HP Laptop 17, 64GB RAM, WD external drives (various drives tried, behavior is the same in all of them)

Please tell us more about your environment

**VeraCrypt version:** 
Veracrypt 1.26.15, 64-bit

**Operating system and version:** 
Windows 11 Pro, 23H3, Build 22631.4169

**System type:** 
64-bit

amail80:
I've investigated this bug and found the root cause. Submitted a fix: #1658 

**Root cause:** `SHChangeNotify(SHCNE_DRIVEREMOVED)` in `BroadcastDeviceChange()` is asynchronous by default. In GUI mode the process stays alive, so Explorer eventually processes the notification. In CLI mode with `/q`, the process exits immediately — before Explorer handles the notification. The drive letter stays in the shell namespace cache.

**Fix:** Add `SHCNF_FLUSH` flag in Silent (CLI) mode to force synchronous shell notification. Two locations: `UnmountVolumeBase()` in `Dlgcode.c` and `DismountAll()` in `Mount.c`.

This also likely fixes the older #337.

DavidOfMA:
Great! Do you have any idea when this fix will be incorporated into a release of Veracrypt?

amail80:
> Great! Do you have any idea when this fix will be incorporated into a release of Veracrypt?

That depends on the maintainer. The PR is ready for review at #1658

idrassi:
Fix merged. Thanks.

Issue #337, opened 2018-07-17
"phantom disk" still present after dismount script from task scheduler


xznhj8129:
I have a script that runs on task scheduler that does a /dismount /force /quit to enable the volume files to be backed up. When tested the script runs fine, however when run from windows Task Scheduler (8.1), the volume letter i was using still shows as used even though it is not accessible or otherwise removable. This makes it impossible to re-use that drive letter until the computer is rebooted (i hope). 
Same problem as these people: https://sourceforge.net/p/veracrypt/discussion/technical/thread/9e67bfd2/

zasifras:
VC 1.24-Update2(64bit) on Win10 still exhibits this bug.   This issue only occurs when using the command line options, for script use.   Using the GUI it works fine and the drive letter is correctly freed up for use again.  But the command line options dismount the disk, but still leave the drive letter unusable in Windows, where any attempted access of that drive letter causes Windows to error.   Only rebooting the Win10 system will free the drive letters.    So it seems there is some difference between the GUI code path and the command line code path when dismounting drives.

rumble-imaging-wages:
I just wanted to report that I'm having similar, though slightly different, problems. While the above and other examples I can see online relate to dismounting drives through scripts / CLI, I'm ending up with "phantom drives" when mounting through the CLI and dismounting through the GUI (right-click taskbar icon -> dismount).

I've discovered a couple of nuances with this behaviour:

1. Problem only occurs if I mount the container using a PowerShell script. Mounting with the exact same parameters using a batch file results in no phantom drives when I dismount.
2.  Problem only occurs for me with encrypted file containers. I can happily mount encrypted non-system partitions using either batch files or PS scripts and successfully dismount them with the GUI leaving no phantom drives behind.

I'm using VC 1.24-Update7 (64bit) on Win 10 Pro.

Batch file command:
```
start "" "C:\Program Files\VeraCrypt\veracrypt.exe" /quit background /letter d /mountoption label=label /volume container_path
```

PowerShell command:
```
$veracrypt = "C:\Program Files\VeraCrypt\veracrypt.exe"
$args = "/quit background", "/letter d", "/mountoption label=label", "/volume container_path"
Start-Process -FilePath $veracrypt -ArgumentList $args
```

alt3r-3go:
[...]

PR #1658, opened 2026-02-15
Fix ghost drive letter after CLI dismount (Silent mode)

amail80:
### Problem

When dismounting volumes via CLI (`VeraCrypt.exe /d X /q /s`), the drive letter remains visible in Windows Explorer as an inaccessible "Local Disk" until reboot. GUI dismount works correctly.

This has been reported multiple times over the past 8 years:

- #337 (2018) — "phantom disk" after Task Scheduler dismount
- #1426 (2024) — CLI dismount doesn't release drive letter

### Root Cause

`BroadcastDeviceChange()` in `Dlgcode.c` calls `SHChangeNotify(SHCNE_DRIVEREMOVED, SHCNF_PATH, ...)` which is **asynchronous** by default. The notification is queued for Explorer to process later.

In GUI mode this works fine — the process stays alive and Explorer eventually processes the notification.

In CLI mode with `/q` (Silent), the process calls `exit()` immediately after dismount (`Mount.c:7575`). Explorer never gets a chance to process the queued notification. The drive letter remains in the shell namespace cache indefinitely.

### Fix

Add `SHCNF_FLUSH` flag to `SHChangeNotify` when operating in Silent (CLI) mode. This forces synchronous processing — the function blocks until Explorer updates its drive list, before the process exits.

Two locations are patched:

- `src/Common/Dlgcode.c` — `UnmountVolumeBase()` (single volume dismount)
- `src/Mount/Mount.c` — `DismountAll()` (dismount all volumes)

The flush is **only added in Silent mode** to avoid introducing latency in interactive GUI operations.

### Testing

This fix should be tested with:

1. `VeraCrypt.exe /d X /q /s` — single volume CLI dismount
2. `VeraCrypt.exe /d /q /s` — dismount all via CLI
3. Task Scheduler automated dismount
4. Verify GUI dismount is unaffected

### DIFF
--- a/src/Common/Dlgcode.c
+++ b/src/Common/Dlgcode.c
@@ -9357,6 +9357,18 @@ retry:
 
 	BroadcastDeviceChange (DBT_DEVICEREMOVECOMPLETE, nDosDriveNo, 0);
 
+	/* GH #337, GH #1426: When running in silent/CLI mode, the process may
+	   exit immediately after unmount. BroadcastDeviceChange sends
+	   SHChangeNotify asynchronously, so Explorer may not process the drive
+	   removal before the process exits, leaving a ghost drive letter.
+	   Re-send the notification with SHCNF_FLUSH to force synchronous
+	   processing by Explorer before we return. */
+	if (Silent)
+	{
+		wchar_t root[] = { (wchar_t) (nDosDriveNo + L'A'), L':', L'\\', 0 };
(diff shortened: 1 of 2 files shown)


idrassi:
Thank for this fix. I missed the asynchronous nature of SHChangeNotify all these years!
The code changes are ok and my tests show that it works properly. 
I will merge it.

PR #1670, opened 2026-04-12
Windows: Fix elevated COM format drive validation and device path normalization

audriusbuika:
This PR addresses two correctness issues discovered while reviewing the
elevated COM surface on Windows. It is intentionally narrow in scope: it
fixes the specific bugs below and does **not** attempt the broader COM
privilege-escalation hardening, which will be handled in a separate
follow-up.

### 1. `Device::Device` — avoid double-prefixing device paths

Previously, `Device::Device` could prepend `\\.\` to a path that already
began with `\\.\`, producing a malformed device path. The constructor now
checks whether the input already starts with `\\.\` and only adds the
prefix when it is missing. The existing `PhysicalDriveN` handling is
preserved unchanged.

### 2. `FormatNtfs` / `FormatFs` — return a non-zero error for invalid drive numbers

Previously, when given an invalid drive number, `FormatNtfs` and
`FormatFs` could return a success status without performing any
formatting. Because `UacFormatNtfs` / `UacFormatFs` forward this status
across the COM boundary, the elevated callers would interpret invalid
input as a successful format. Both functions now return a non-zero error
code for invalid drive numbers, so the UAC wrappers correctly propagate
the failure to the caller.

## Scope

This PR is limited to the two correctness fixes above. The broader
hardening of the elevated COM privilege-escalation surface that
originally motivated this review is deferred to a follow-up PR.

### DIFF
--- a/src/Common/BaseCom.cpp
+++ b/src/Common/BaseCom.cpp
@@ -21,6 +21,155 @@
 
 using namespace VeraCrypt;
 
+// ========================================================================
+// COM method input validation - whitelists and helper functions
+// ========================================================================
+
+// Helper: check if IOCTL code is in the allowed list
+static BOOL IsIoctlInWhitelist (DWORD ioctl, const DWORD *whitelist, size_t count)
+{
+	for (size_t i = 0; i < count; i++)
+	{
+		if (whitelist[i] == ioctl)
+			return TRUE;
+	}
+	return FALSE;
+}
+
+// Allowed IOCTL codes for CallDriver (VeraCrypt kernel driver IOCTLs)
+static const DWORD g_CallDriverIoctlWhitelist[] = {
+	TC_IOCTL_GET_DRIVER_VERSION,
+	TC_IOCTL_GET_BOOT_LOADER_VERSION,
+	TC_IOCTL_MOUNT_VOLUME,
+	TC_IOCTL_UNMOUNT_VOLUME,
+	TC_IOCTL_UNMOUNT_ALL_VOLUMES,
+	TC_IOCTL_GET_MOUNTED_VOLUMES,
+	TC_IOCTL_GET_VOLUME_PROPERTIES,
+	TC_IOCTL_GET_DEVICE_REFCOUNT,
+	TC_IOCTL_IS_DRIVER_UNLOAD_DISABLED,
+	TC_IOCTL_IS_ANY_VOLUME_MOUNTED,
+	TC_IOCTL_GET_PASSWORD_CACHE_STATUS,
+	TC_IOCTL_WIPE_PASSWORD_CACHE,
+	TC_IOCTL_OPEN_TEST,
+	TC_IOCTL_GET_DRIVE_PARTITION_INFO,
+	TC_IOCTL_GET_DRIVE_GEOMETRY,
+	TC_IOCTL_PROBE_REAL_DRIVE_SIZE,
+	TC_IOCTL_GET_RESOLVED_SYMLINK,
+	TC_IOCTL_GET_BOOT_ENCRYPTION_STATUS,
+	TC_IOCTL_BOOT_ENCRYPTION_SETUP,
+	TC_IOCTL_ABORT_BOOT_ENCRYPTION_SETUP,
+	TC_IOCTL_GET_BOOT_ENCRYPTION_SETUP_RESULT,
+	TC_IOCTL_GET_BOOT_DRIVE_VOLUME_PROPERTIES,
+	TC_IOCTL_REOPEN_BOOT_VOLUME_HEADER,
+	TC_IOCTL_GET_BOOT_ENCRYPTION_ALGORITHM_NAME,
+	TC_IOCTL_GET_PORTABLE_MODE_STATUS,
+	TC_IOCTL_SET_PORTABLE_MODE_STATUS,
+	TC_IOCTL_IS_HIDDEN_SYSTEM_RUNNING,
+	TC_IOCTL_GET_SYSTEM_DRIVE_CONFIG,
+	TC_IOCTL_DISK_IS_WRITABLE,
+	TC_IOCTL_START_DECOY_SYSTEM_WIPE,
+	TC_IOCTL_ABORT_DECOY_SYSTEM_WIPE,
+	TC_IOCTL_GET_DECOY_SYSTEM_WIPE_STATUS,
+	TC_IOCTL_GET_DECOY_SYSTEM_WIPE_RESULT,
+	TC_IOCTL_WRITE_BOOT_DRIVE_SECTOR,
+	TC_IOCTL_GET_WARNING_FLAGS,
+	TC_IOCTL_SET_SYSTEM_FAVORITE_VOLUME_DIRTY,
+	TC_IOCTL_REREAD_DRIVER_CONFIG,
+	TC_IOCTL_GET_SYSTEM_DRIVE_DUMP_CONFIG,
+	VC_IOCTL_GET_BOOT_LOADER_FINGERPRINT,
+	VC_IOCTL_GET_DRIVE_GEOMETRY_EX,
+	VC_IOCTL_EMERGENCY_CLEAR_ALL_KEYS,
+	VC_IOCTL_IS_RAM_ENCRYPTION_ENABLED,
+	VC_IOCTL_ENCRYPTION_QUEUE_PARAMS,
+};
+
+// Allowed IOCTL codes for DeviceIoControl (standard disk/storage/filesystem IOCTLs)
+static const DWORD g_DeviceIoControlWhitelist[] = {
+	IOCTL_DISK_GET_DRIVE_GEOMETRY,
+	IOCTL_DISK_GET_PARTITION_INFO,
+	IOCTL_DISK_GET_PARTITION_INFO_EX,
+	IOCTL_DISK_GET_DRIVE_GEOMETRY_EX,
+	IOCTL_DISK_GET_LENGTH_INFO,
+	IOCTL_DISK_PERFORMANCE,
+	IOCTL_STORAGE_GET_DEVICE_NUMBER,
+	IOCTL_STORAGE_READ_CAPACITY,
+	IOCTL_STORAGE_MANAGE_DATA_SET_ATTRIBUTES,
+	FSCTL_LOCK_VOLUME,
+	FSCTL_DISMOUNT_VOLUME,
+	FSCTL_IS_VOLUME_MOUNTED,
+	FSCTL_GET_NTFS_VOLUME_DATA,
+	FSCTL_GET_VOLUME_BITMAP,
+	FSCTL_GET_RETRIEVAL_POINTERS,
+	FSCTL_MOVE_FILE,
+	FSCTL_ALLOW_EXTENDED_DASD_IO,
+	FSCTL_SET_SPARSE,
+	FSCTL_EXTEND_VOLUME,
+	FSCTL_SHRINK_VOLUME,
+};
+
+// Allowed HKLM registry key path prefixes (case-insensitive)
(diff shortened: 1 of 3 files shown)


idrassi:
Thank you for looking into this area.
I agree this COM surface needs hardening but this patch cannot be merged as-is. It appears to introduce at least two correctness regressions: device paths such as PhysicalDriveN used by boot encryption fallback are rejected before the elevated Device wrapper can open them, and invalid FormatNtfs/FormatFs drive numbers return FALSE/0, which the existing caller treats as success.

The validation policy also needs to be narrower. The driver IOCTL whitelist currently includes all current VeraCrypt driver IOCTLs, including mutating operations, so it does not materially constrain the elevated broker beyond blocking unknown codes. The DeviceIoControl list also includes mutating FSCTLs that do not appear to be required by current elevated COM call sites. For path-based methods, substring checks like "contains veracrypt" and .. rejection are not a reliable location policy.

I think this should be reworked incrementally: first preserve existing elevated boot/system drive behavior, fix the format return value bug, and derive whitelists from actual elevated COM call sites. Then add per-operation validation where practical, with special treatment for workflows that genuinely require caller-provided paths, such as fast creation of user selected containers.

src/Common/BaseCom.cpp
```diff
@@ -21,6 +21,155 @@
...
+	TC_IOCTL_GET_PORTABLE_MODE_STATUS,
+	TC_IOCTL_SET_PORTABLE_MODE_STATUS,
+	TC_IOCTL_IS_HIDDEN_SYSTEM_RUNNING,
+	TC_IOCTL_GET_SYSTEM_DRIVE_CONFIG,
+	TC_IOCTL_DISK_IS_WRITABLE,
+	TC_IOCTL_START_DECOY_SYSTEM_WIPE,
+	TC_IOCTL_ABORT_DECOY_SYSTEM_WIPE,
+	TC_IOCTL_GET_DECOY_SYSTEM_WIPE_STATUS,
+	TC_IOCTL_GET_DECOY_SYSTEM_WIPE_RESULT,
+	TC_IOCTL_WRITE_BOOT_DRIVE_SECTOR,
```
idrassi: This whitelist currently includes every current driver IOCTL, including mutating or destructive operations such as TC_IOCTL_BOOT_ENCRYPTION_SETUP, TC_IOCTL_WRITE_BOOT_DRIVE_SECTOR, TC_IOCTL_START_DECOY_SYSTEM_WIPE and VC_IOCTL_EMERGENCY_CLEAR_ALL_KEYS.

That does not create a meaningful authorization boundary for the elevated COM server. It only blocks unknown future IOCTLs. Please narrow this to IOCTLs actually needed through elevated COM and add per-IOCTL validation for input/output sizes, structure fields, target device, offsets, and setup state where applicable before dispatching.

src/Common/BaseCom.cpp
```diff
@@ -21,6 +21,155 @@
...
+	FSCTL_LOCK_VOLUME,
+	FSCTL_DISMOUNT_VOLUME,
+	FSCTL_IS_VOLUME_MOUNTED,
+	FSCTL_GET_NTFS_VOLUME_DATA,
+	FSCTL_GET_VOLUME_BITMAP,
+	FSCTL_GET_RETRIEVAL_POINTERS,
+	FSCTL_MOVE_FILE,
+	FSCTL_ALLOW_EXTENDED_DASD_IO,
+	FSCTL_SET_SPARSE,
+	FSCTL_EXTEND_VOLUME,
```
idrassi: The mutating FSCTLs FSCTL_MOVE_FILE / FSCTL_EXTEND_VOLUME should not be exposed through a generic elevated DeviceIoControl broker without operation-specific validation. With this list, the caller still controls both the target path/device and the input buffer for operations such as FSCTL_MOVE_FILE, FSCTL_EXTEND_VOLUME, FSCTL_SHRINK_VOLUME and
IOCTL_STORAGE_MANAGE_DATA_SET_ATTRIBUTES.

Please remove IOCTLs that are not required by current elevated COM call sites or replace them with dedicated broker methods that validate the exact target, buffer size, fields, and operation context.

src/Common/BaseCom.cpp
```diff
@@ -21,6 +21,155 @@
...
+	}
+	return FALSE;
+}
+
+// Helper: reject paths containing ".." traversal components
+static BOOL ValidatePathNoTraversal (const wchar_t *path)
+{
+	if (!path)
+		return FALSE;
+	if (wcsstr (path, L"..") != NULL)
```
idrassi: This is not a safe path validation policy. It rejects legitimate file names containing "..." but still accepts arbitrary absolute paths such as "C:\Windows\..." and it does not handle canonicalization, junctions, symlinks, mount points, or other reparse points.

Please replace this with canonical path validation tied to each operation's expected target. User selected container paths should be treated as a separate workflow with explicit validation.

src/Common/BaseCom.cpp
```diff
@@ -21,6 +21,155 @@
...
+static BOOL PathContainsVeraCrypt (const wchar_t *path)
+{
+	if (!path)
+		return FALSE;
+	size_t pathLen = wcslen (path);
+	if (pathLen < 9)
+		return FALSE;
+	for (size_t i = 0; i <= pathLen - 9; i++)
+	{
+		if (_wcsnicmp (&path[i], L"veracrypt", 9) == 0)
```
idrassi: Checking for a veracrypt substring is not a safe location policy. A caller can choose arbitrary paths containing that substring and this does not protect against reparse-point redirection.

Please restrict CopyFile/DeleteFile to the exact expected VeraCrypt-managed files after resolving/canonicalizing the path or replace these generic methods with dedicated operations for the known copy/delete workflows.

src/Common/BaseCom.cpp
```diff
@@ -21,6 +21,155 @@
...
+	}
+	return FALSE;
+}
+
+// Helper: validate device path format (must start with \\.\ or \\?\)
+static BOOL IsValidDevicePath (const wchar_t *path)
+{
+	if (!path || wcslen (path) < 4)
+		return FALSE;
+	return (path[0] == L'\\' && path[1] == L'\\' && (path[2] == L'.' || path[2] == L'?') && path[3] == L'\\');
```
idrassi: This breaks existing elevated device fallback paths. Device::Device opens PhysicalDriveN locally by prepending \\.\ but stores the original Path. System-drive configuration uses values like PhysicalDrive. When non-elevated access fails with ERROR_ACCESS_DENIED, Elevator::ReadWriteFile / Elevator::DeviceIoControl passes that original PhysicalDrive0 string to COM and this validator rejects it.

Accepting \\.\PhysicalDrive0 here is not sufficient by itself. Device::Device only preserves paths that already start with \\?\. Other inputs are prefixed with "\\.\" so passing \\.\PhysicalDrive0 into Device::Device would produce an invalid doubled prefix. Please normalize device paths consistently in one place while preserving the existing PhysicalDriveN elevated fallback behavior.

src/Common/BaseCom.cpp
```diff
@@ -134,6 +296,17 @@ BOOL BaseCom::IsPagingFileActive (BOOL checkNonWindowsPartitionsOnly)
 
 DWORD BaseCom::ReadWriteFile (BOOL write, BOOL device, BSTR filePath, BSTR *bufferBstr, unsigned __int64 offset, unsigned __int32 size, DWORD *sizeDone)
 {
+	if (device)
+	{
+		if (!IsValidDevicePath (filePath))
```
idrassi: For device == TRUE, this currently performs only syntactic prefix validation. Even after the path normalization issue is fixed, this remains a generic elevated raw device read/write primitive.

Please constrain elevated device reads/writes to the boot/system-drive workflows that require them and validate the target device plus permitted offsets and sizes for each operation.

src/Common/BaseCom.cpp
```diff
@@ -302,6 +492,9 @@ DWORD BaseCom::SetDriverServiceStartType (DWORD startType)
 
 DWORD BaseCom::WriteLocalMachineRegistryDwordValue (BSTR keyPath, BSTR valueName, DWORD value)
 {
+	if (!IsRegistryKeyPathAllowed (keyPath))
```
idrassi: This prefix whitelist is still too broad for a generic elevated DWORD write method. Current callers only need exact writes such as `veracrypt\VeraCryptConfig` with known config bits, `VeraCryptEnableMemoryProtection`/`VeraCryptEnableScreenProtection` as 0 or 1, `VeraCryptEncryptionFreeCpuCount` within the CPU-count limit, `VeraCryptSystemFavorites\VeraCryptSystemFavoritesConfig` with known service-config bits and `HibernateEnabled`/`HiberbootEnabled` set to 0.

Please validate exact `(keyPath, valueName, value)` combinations. For example, this method should not allow unrelated service values such as `SYSTEM\CurrentControlSet\Services\veracrypt\Start`: service start changes already have the dedicated `SetDriverServiceStartType` path.

src/Common/BaseCom.cpp
```diff
@@ -500,5 +693,8 @@ DWORD BaseCom::NotifyService(DWORD dwNotifyCode)
 
 DWORD BaseCom::FastFileResize (BSTR filePath, __int64 fileSize)
 {
+	if (!ValidatePathNoTraversal (filePath))
```
idrassi: FastFileResize is especially sensitive because it enables SE_MANAGE_VOLUME_NAME and calls SetFileValidData. A ".." substring check is not enough for a privileged operation on a caller-provided path.

Please validate that the target is the container file created by the formatting workflow, reject reparse-point/symlink redirection and avoid accepting arbitrary absolute paths through this elevated method.

src/Format/FormatCom.cpp
```diff
@@ -92,6 +92,8 @@ class TrueCryptFormatCom : public ITrueCryptFormatCom
 
 	virtual BOOL STDMETHODCALLTYPE FormatNtfs (int driveNo, int clusterSize)
 	{
+		if (driveNo < 0 || driveNo > 25)
+			return FALSE;
```
idrassi: This returns FALSE but UacFormatNtfs consumes the COM return value as an int status where 0 means success and non-zero means failure. Invalid input would therefore be reported as a successful format.

Please return a non-zero failure code such as ERROR_INVALID_DRIVE or ERROR_INVALID_PARAMETER or update the COM method contract consistently so invalid input
cannot be interpreted as success.

src/Format/FormatCom.cpp
```diff
@@ -134,6 +136,8 @@ class TrueCryptFormatCom : public ITrueCryptFormatCom
 
 	virtual BOOL STDMETHODCALLTYPE FormatFs (int driveNo, int clusterSize, int fsType)
 	{
+		if (driveNo < 0 || driveNo > 25)
+			return FALSE;
```
idrassi: Same issue as FormatNtfs: FALSE is 0 and the caller treats 0 as success. Please return a non-zero error code for invalid drive numbers.

idrassi (changes requested):

idrassi:
Thank you for reworking this.

The current final diff is much safer than the previous version.

The remaining changes look reasonable to me as a small correctness fix:
- Device::Device now avoids double-prefixing paths that already start with \\.\, while preserving the existing PhysicalDriveN behavior.
- FormatNtfs/FormatFs now return a non-zero error for invalid drive numbers, so UacFormatNtfs/UacFormatFs will no longer interpret invalid input as success.

I would only ask that the PR be re-scoped to match what it now does. In its current state, it no longer fixes the broader elevated COM privilege-escalation surface: it fixes two correctness issues discovered while reviewing that area.

A title such as "Windows: Fix elevated COM format drive validation and device path normalization" would be more accurate, with the broader COM hardening deferred to a follow-up work.

Can you please do this?

idrassi (approved):

Issue #1668, opened 2026-04-09
Fresh windows 11, hibernate = crash instead of hibernation


kstepyra:
As in title - instead of hibernating it crashes and requires boot password - event viewer = critical err 41 (no info why it rebooted).

# Expected behavior
work

# Observed behavior
not working..

# Steps to reproduce
1. install windows 11
2. encrypt via VC
3. enable hibernation
4. hibernate and crash

# Your Environment

**VeraCrypt version:**: newest version as of today - 1.26.24

**Operating system and version:** Windows 11 pro 25H2

**System type:** 64-bit

idrassi:
Thank you for reporting this issue.

PR #1671 has been opened to address this hibernation crash and is currently under review.

The proposed core fix is in the Windows dump/hibernate filter path. Review comments have been left on the PR and we are waiting for the contributor’s feedback/updates before it can be considered for merge.

PR #1671, opened 2026-04-12
Fix hibernation crash on fresh Windows 11 25H2 (BSOD Event 41)

audriusbuika:
solved problem: #1668

### DIFF
--- a/src/Driver/DriveFilter.c
+++ b/src/Driver/DriveFilter.c
@@ -968,12 +968,7 @@ static NTSTATUS DispatchPower (PDEVICE_OBJECT DeviceObject, PIRP Irp, DriveFilte
 		&& irpSp->MinorFunction == IRP_MN_SET_POWER
 		&& irpSp->Parameters.Power.ShutdownType == PowerActionHibernate)
 	{
-		int retries = 50;
-		while (retries-- > 0
-			&& SendDeviceIoControlRequest (RootDeviceObject, TC_IOCTL_ABORT_BOOT_ENCRYPTION_SETUP, NULL, 0, NULL, 0) == STATUS_INSUFFICIENT_RESOURCES)
-		{
-			TCSleep (100);
-		}
+		while (SendDeviceIoControlRequest (RootDeviceObject, TC_IOCTL_ABORT_BOOT_ENCRYPTION_SETUP, NULL, 0, NULL, 0) == STATUS_INSUFFICIENT_RESOURCES);
 	}
 
 	// Dismount the system drive on shutdown on Windows 7 and later
@@ -1212,7 +1207,7 @@ wipe:
 /*
  * Legacy Windows XP/2003 hibernation dump filter.
  *
- * DISABLED: This code is not active — LoadImageNotifyRoutine was never
+ * DISABLED: This code is not active - LoadImageNotifyRoutine was never
  * registered via PsSetLoadImageNotifyRoutine, so none of these functions
  * are reachable at runtime.  Additionally the code has known issues:
  *   - HibernationWriteBuffer / HibernationWriteBufferMdl are never allocated
--- a/src/Driver/DriveFilter.c
+++ b/src/Driver/DriveFilter.c
@@ -52,10 +52,6 @@ static uint8 BootLoaderFingerprint[WHIRLPOOL_DIGESTSIZE + SHA512_DIGESTSIZE];
 static BOOL CrashDumpEnabled = FALSE;
 static BOOL HibernationEnabled = FALSE;
 
-static BOOL LegacyHibernationDriverFilterActive = FALSE;
-static uint8 *HibernationWriteBuffer = NULL;
-static MDL *HibernationWriteBufferMdl = NULL;
-
 static uint32 HibernationPreventionCount = 0;
 
 static BootEncryptionSetupRequest SetupRequest;
@@ -972,7 +968,12 @@ static NTSTATUS DispatchPower (PDEVICE_OBJECT DeviceObject, PIRP Irp, DriveFilte
 		&& irpSp->MinorFunction == IRP_MN_SET_POWER
 		&& irpSp->Parameters.Power.ShutdownType == PowerActionHibernate)
 	{
-		while (SendDeviceIoControlRequest (RootDeviceObject, TC_IOCTL_ABORT_BOOT_ENCRYPTION_SETUP, NULL, 0, NULL, 0) == STATUS_INSUFFICIENT_RESOURCES);
+		int retries = 50;
+		while (retries-- > 0
+			&& SendDeviceIoControlRequest (RootDeviceObject, TC_IOCTL_ABORT_BOOT_ENCRYPTION_SETUP, NULL, 0, NULL, 0) == STATUS_INSUFFICIENT_RESOURCES)
+		{
+			TCSleep (100);
+		}
 	}
 
 	// Dismount the system drive on shutdown on Windows 7 and later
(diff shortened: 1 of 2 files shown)


idrassi:
@audriusbuika 
Thanks for working on this. I reviewed the patch, and the `DumpFilter.c` change looks like the core fix here.

Some remarks:

1. The `DriveFilter.c` retry/sleep change should either be split out or tightened. I traced the local call path into this branch and `DispatchPower` is reached directly from the incoming `IRP_MJ_POWER` dispatch path. There is no worker-item handoff or local guarantee that this runs at `PASSIVE_LEVEL`. That matters because `TCSleep` uses `KeDelayExecutionThread` and requires `IRQL <= APC_LEVEL`. For hibernation-file device paths, Windows requires drivers to lock to run at `PASSIVE_LEVEL`.

2. The bounded retry changes an important invariant in this path. `STATUS_INSUFFICIENT_RESOURCES` from `SendDeviceIoControlRequest` can mean the driver failed to allocate/queue/build the internal IOCTL request, not that `AbortBootEncryptionSetup` ran and failed. If all 50 attempts return `STATUS_INSUFFICIENT_RESOURCES`, the abort request may never have reached `AbortBootEncryptionSetup`, so `EncryptionSetupThreadAbortRequested` may remain false and the setup thread may still be active. That conflicts with the invariant already documented in `DumpFilterWrite` (`EncryptedAreaEndUpdatePending` -> "Hibernation should always abort the setup thread"). Please make the post-retry behavior preserve that invariant, for example by proving setup has actually stopped before continuing, or by handling/logging the failure path explicitly instead of silently proceeding as if the abort completed.
3. The direct `KeBugCheckEx` call uses the same `SECURITY_SYSTEM` bugcheck code and argument convention as `TC_BUG_CHECK`. The only difference is that it puts `writeMdl->MdlFlags` in bugcheck parameter 3 instead of `0`. That is useful diagnostically, but it does bypass the local `TC_BUG_CHECK` convention. This could either use `TC_BUG_CHECK(STATUS_INVALID_PARAMETER)` and drop the extra diagnostic value, or introduce a small `TC_BUG_CHECK_EX(status, detail)` helper.

Minor style note: the new comments contain Unicode punctuation (em-dash). Using ASCII hyphens/apostrophes would avoid GitHub's hidden/Unicode warning noise.

My recommendation would be to keep the `DumpFilter.c` MDL fix as the core bug fix but rework or split out the `DispatchPower` retry/sleep change until its IRQL assumptions and "setup stopped before hibernation" invariant are explicit.

src/Driver/DriveFilter.c
```diff
@@ -972,7 +968,12 @@ static NTSTATUS DispatchPower (PDEVICE_OBJECT DeviceObject, PIRP Irp, DriveFilte
 		&& irpSp->MinorFunction == IRP_MN_SET_POWER
 		&& irpSp->Parameters.Power.ShutdownType == PowerActionHibernate)
 	{
-		while (SendDeviceIoControlRequest (RootDeviceObject, TC_IOCTL_ABORT_BOOT_ENCRYPTION_SETUP, NULL, 0, NULL, 0) == STATUS_INSUFFICIENT_RESOURCES);
+		int retries = 50;
+		while (retries-- > 0
+			&& SendDeviceIoControlRequest (RootDeviceObject, TC_IOCTL_ABORT_BOOT_ENCRYPTION_SETUP, NULL, 0, NULL, 0) == STATUS_INSUFFICIENT_RESOURCES)
+		{
+			TCSleep (100);
```
idrassi: Please avoid sleeping directly in this `DispatchPower` path unless the IRQL guarantee is explicit. This branch is reached directly from the incoming `IRP_MJ_POWER` dispatch path. I do not see a local worker-item handoff or other guarantee that it runs at `PASSIVE_LEVEL`. `TCSleep` wraps
`KeDelayExecutionThread` and requires `IRQL <= APC_LEVEL`. For hibernation-related power paths, the stack may have cleared `DO_POWER_PAGABLE`, so this needs either an explicit IRQL-safe implementation or a move to a path that is guaranteed to run at a waitable IRQL.

src/Driver/DriveFilter.c
```diff
@@ -972,7 +968,12 @@ static NTSTATUS DispatchPower (PDEVICE_OBJECT DeviceObject, PIRP Irp, DriveFilte
 		&& irpSp->MinorFunction == IRP_MN_SET_POWER
 		&& irpSp->Parameters.Power.ShutdownType == PowerActionHibernate)
 	{
-		while (SendDeviceIoControlRequest (RootDeviceObject, TC_IOCTL_ABORT_BOOT_ENCRYPTION_SETUP, NULL, 0, NULL, 0) == STATUS_INSUFFICIENT_RESOURCES);
+		int retries = 50;
+		while (retries-- > 0
+			&& SendDeviceIoControlRequest (RootDeviceObject, TC_IOCTL_ABORT_BOOT_ENCRYPTION_SETUP, NULL, 0, NULL, 0) == STATUS_INSUFFICIENT_RESOURCES)
```
idrassi: This bounded retry can leave boot encryption setup running. `STATUS_INSUFFICIENT_RESOURCES` from `SendDeviceIoControlRequest()` can be returned before the IOCTL reaches `AbortBootEncryptionSetup()` at all, for example if the work item or internal IOCTL IRP cannot be allocated. After 50 attempts this code falls through even though `EncryptionSetupThreadAbortRequested` may still be false and `SetupInProgress` may still be true. That conflicts with the invariant already documented in `DumpFilterWrite` that hibernation should always abort the setup thread before dump writes proceed. Please preserve that invariant by proving the setup thread has stopped before continuing, or make the failure path explicit instead of silently proceeding as if the abort completed.

src/Driver/DumpFilter.c
```diff
@@ -266,11 +266,26 @@ static NTSTATUS DumpFilterWrite (PFILTER_EXTENSION filterExtension, PLARGE_INTEG
...
 
-	// Require either a valid mapping or a nonpaged system VA we can read at HIGH_LEVEL.
-	if ((writeMdl->MdlFlags & (MDL_MAPPED_TO_SYSTEM_VA | MDL_SOURCE_IS_NONPAGED_POOL)) == 0)
-		TC_BUG_CHECK(STATUS_INVALID_PARAMETER);
+	// Resolve the system VA we can safely read at HIGH_LEVEL.
+	// MDL_SOURCE_IS_NONPAGED_POOL: original VA (StartVa + ByteOffset) is a kernel VA.
+	// MDL_MAPPED_TO_SYSTEM_VA:     MappedSystemVa is the correct kernel VA (StartVa
+	//                              may point elsewhere, e.g. user-mode or stale VA).
+	// Windows 11 25H2+ dump stacks may provide mapped MDLs that are NOT from nonpaged
+	// pool, so we must prefer MappedSystemVa when available.
```
idrassi, line 274: Small clarification: this comment says we must prefer `MappedSystemVa` when available, but the code below gives `MDL_SOURCE_IS_NONPAGED_POOL` precedence over `MDL_MAPPED_TO_SYSTEM_VA`. If that precedence is intentional, please adjust the comment so it matches the implementation. If `MappedSystemVa` should really win whenever `MDL_MAPPED_TO_SYSTEM_VA` is set, the checks should be reordered.

idrassi (changes requested):
The `DumpFilter.c` MDL address selection looks like the right core fix for the hibernation crash. I am requesting changes because the `DispatchPower` retry/sleep change introduces separate power-path concerns: it may sleep without a local IRQL guarantee and the bounded retry can proceed without proving that boot encryption setup was actually aborted. Please either rework that part to preserve the existing hibernation/setup invariant safely, or split it out so the MDL fix can be reviewed independently.

idrassi (approved):

idrassi:
@audriusbuika Thank you for addressing the comments. Proceeding with the merge.

ghost:
"_Minor style note: the new comments **contain Unicode punctuation (em-dash)**. Using ASCII hyphens/apostrophes would avoid GitHub's hidden/Unicode warning noise._"

This evidently suggests that @audriusbuika used AI for coding; almost no one use em-dash in coding if they use hands and keyboard.

@idrassi Do you accept AI-assited code now? Be careful when you merge unknown code as it may create hidden vulnerability.

idrassi:
@sranders1 
Yes, I know it is AI but I focus on the quality on the change. That's why I rejected the first version of the PR and asked for correction. I never merge code as-is, being it from human or AI!

There are other PRs that have been submitted and which are still not merged because of various issues I put on comments and their author still didn't address them. Some of them are clearly AI with a lot of slop.

Humans can also propose malicious changes/PRs so I always follow the same approach: deep evaluation and always suspicious. AI just increase volume but honestly I'm not (yet) drowning under AI slot.

PR #1672, opened 2026-04-12
Fix erroneous 2 TiB limit for hidden file containers in GUI wizard

### LINKED ISSUE #1560: Unable to create > 2TB hidden volume in a file container on Linux using GUI wizard
# Summary
In the GUI wizard
1. `SectorSize` is set to `TC_SECTOR_SIZE_FILE_HOSTED_VOLUME` ([VolumeCreationWizard.cpp:701](https://github.com/veracrypt/VeraCrypt/blob/4402d153b9a94785d9f70ee3b377466ea08f6d23/src/Main/Forms/VolumeCreationWizard.cpp#L70))
2. There is a check for `VolumeSize > TC_MAX_FAT_SECTOR_COUNT * SectorSize` ([VolumeCreationWizard.cpp:976](https://github.com/veracrypt/VeraCrypt/blob/4402d153b9a94785d9f70ee3b377466ea08f6d23/src/Main/Forms/VolumeCreationWizard.cpp#L976))
3. `TC_SECTOR_SIZE_FILE_HOSTED_VOLUME = 512`
4. `TC_MAX_FAT_SECTOR_COUNT = 0x100000000ULL`
5. This result in a cap of 2TiB when creating an outer container since `TC_SECTOR_SIZE_FILE_HOSTED_VOLUME * TC_MAX_FAT_SECTOR_COUNT = 2TiB`

# Recommendation
Change [VolumeCreationWizard.cpp:976](https://github.com/veracrypt/VeraCrypt/blob/4402d153b9a94785d9f70ee3b377466ea08f6d23/src/Main/Forms/VolumeCreationWizard.cpp#L976) to read: `if (SelectedVolumePath.IsDevice() && OuterVolume && VolumeSize > TC_MAX_FAT_SECTOR_COUNT * SectorSize)` since the sector check is irrelevant in non-gui applications
Also this restriction can by bypassed using the command line as noticed by: https://sourceforge.net/p/veracrypt/discussion/technical/thread/383b4bedb0/

If this is acceptable, I can create a PR, please let me know.

# Steps to reproduce
1. Open Veracrypt GUI
2. Create Volume
3. Create an encrypted file container -> Next
4. Hidden VeraCrypt Volume -> Next
5. Select a file on disk
6. Encryption Settings -> Next
7. Volume Size: 3TB
8. Set a password
9. Volume format: Any
10. I will mount only on Linux
11. Format

# Your Environment

**VeraCrypt version:** 1.26.24

**Operating system and version:** Linux

**System type:** 64 bit

Computeiful:
This PR fixes: https://github.com/veracrypt/VeraCrypt/issues/1560

## Summary

Fix the GUI wizard incorrectly rejecting hidden file containers larger than 2 TiB.

Currently the outer hidden-volume size check is applied even when the selected target is a normal file. In that case the wizard uses the file-hosted sector size constant, so the FAT sector-count check incorrectly caps the GUI flow at 2 TiB.

This change restricts that check to device-hosted paths only:

```cpp
if (SelectedVolumePath.IsDevice() && OuterVolume && VolumeSize > TC_MAX_FAT_SECTOR_COUNT * SectorSize)
```

This has been tested on Linux using a 3 TiB file.

### DIFF
--- a/src/Main/Forms/VolumeCreationWizard.cpp
+++ b/src/Main/Forms/VolumeCreationWizard.cpp
@@ -973,7 +973,7 @@ namespace VeraCrypt
 				{
 					if (SelectedVolumeType != VolumeType::Hidden || OuterVolume)
 					{
-						if (OuterVolume && VolumeSize > TC_MAX_FAT_SECTOR_COUNT * SectorSize)
+						if (SelectedVolumePath.IsDevice() && OuterVolume && VolumeSize > TC_MAX_FAT_SECTOR_COUNT * SectorSize)
 						{
 							uint64 limit = TC_MAX_FAT_SECTOR_COUNT * SectorSize / BYTES_PER_TB;
 							wstring err = static_cast<wstring>(StringFormatter (LangString["LINUX_ERROR_SIZE_HIDDEN_VOL"], limit, limit * 1024));


idrassi:
Thank you for the fix.

PR #1675, opened 2026-04-15
Linux/FUSE: tune FUSE 3 init config and fix error mapping

abhinavagarwal07:
Two small fixes for the FUSE layer, building on cdc00dc4.

1. Configure FUSE 3 caching in init

VeraCrypt exposes a fixed three-entry filesystem (`/`, `/volume`, `/control`) that never changes during a mount. With the default 1s `entry_timeout`, the kernel revalidates every dentry once per second — unnecessary round-trips to userspace that add up since the loop device and upper filesystem trigger frequent lookups.

2. Return EIO instead of EINTR for unhandled exceptions

`ExceptionToErrorCode()` was returning EINTR for unknown exceptions. EINTR means "interrupted by signal, retry".

### DIFF
--- a/src/Driver/Fuse/FuseService.cpp
+++ b/src/Driver/Fuse/FuseService.cpp
@@ -125,6 +125,19 @@ namespace VeraCrypt
 			cfg->set_gid = 1;
 			cfg->uid = FuseService::GetUserId();
 			cfg->gid = FuseService::GetGroupId();
+
+			// Honor the inode numbers assigned in getattr/readdir responses.
+			// Without this the kernel may synthesize its own, making the
+			// explicit VC_FUSE_INODE_* values unused.
+			cfg->use_ino = 1;
+
+			// VeraCrypt exposes a fixed three-entry filesystem ("/",
+			// "/volume", "/control") whose structure never changes for the
+			// lifetime of the mount.  Cache lookups aggressively so the
+			// kernel does not round-trip to userspace for every stat/lookup.
+			cfg->entry_timeout = 86400;
+			cfg->negative_timeout = 86400;
+			cfg->attr_timeout = 1;
 		}
 
 		return fuse_service_init_common ();
--- a/src/Driver/Fuse/FuseService.cpp
+++ b/src/Driver/Fuse/FuseService.cpp
@@ -126,18 +126,9 @@ namespace VeraCrypt
 			cfg->uid = FuseService::GetUserId();
 			cfg->gid = FuseService::GetGroupId();
 
-			// Honor the inode numbers assigned in getattr/readdir responses.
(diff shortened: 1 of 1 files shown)


idrassi:
Thank you for the PR.

I reviewed the change and I don't think this can be merged as-is.

The `use_ino = 1` change looks reasonable: VeraCrypt now provides stable inode numbers for the fixed auxiliary FUSE entries and this makes FUSE3 honor them for `stat`/`readdir`.

The `ExceptionToErrorCode` change from `-EINTR` to `-EIO` for otherwise-unhandled exceptions also looks correct.

The problematic part is:

```cpp
cfg->entry_timeout = 86400.0;
```

VeraCrypt auxiliary FUSE mount uses allow_other on Linux and relies on per-request fuse_get_context()->uid checks in userspace.
libfuse documents a known security issue with allow_other when default_permissions is not used: permission-check results for a directory entry may be reused while the inode remains cached, including for later accesses by another user. Increasing positive entry caching to 24 hours expands that risk. See https://github.com/libfuse/libfuse/blob/master/README.md#security-implications and https://libfuse.github.io/doxygen/structfuse__config.html .

For this reason, please remove the long entry_timeout change or rework the mount to use default_permissions and provide evidence that the existing mode bits and access model remain correct for VeraCrypt auxiliary mount. I would also avoid changing permission-sensitive lookup caching unless we have a runtime test covering cross-user access behavior with allow_other.

There is also a smaller issue: the commit/PR description still says negative_timeout is set to 86400 seconds, but the final code no longer sets negative_timeout. Please update the description/commit message so it matches the submitted code.

src/Driver/Fuse/FuseService.cpp
```diff
@@ -125,6 +125,10 @@ namespace VeraCrypt
 			cfg->set_gid = 1;
 			cfg->uid = FuseService::GetUserId();
 			cfg->gid = FuseService::GetGroupId();
+
+			cfg->use_ino = 1;
+			cfg->entry_timeout = 86400.0;
```
idrassi, line 130: VeraCrypt auxiliary FUSE mount uses `allow_other` on Linux and relies on userspace `fuse_get_context()->uid` checks for access control. libfuse documents a known security issue with `allow_other` when `default_permissions` is not used: permission-check results for cached directory entries may be reused for later accesses, including accesses by another user, while the inode remains in kernel cache.

Setting `entry_timeout` to 24 hours expands that risk for a security-sensitive mount. Please either remove this `entry_timeout` change or rework the mount to use `default_permissions` and provide validation that cross-user access behavior remains correct.

idrassi (changes requested):

idrassi:
@abhinavagarwal07 closing this PR because of lack of feedback and the narrowed valid scope

PR #1676, opened 2026-04-15
Linux: fix suspend deadlock with mounted volumes on kernel 6.18+

### LINKED ISSUE #1660: Kernel 6.18.x and 6.19.0-rc5 fail to suspend (hang) if any VeraCrypt volume is mounted
Laptop fails to enter Suspend mode with all versions of Kernel 6.18 and Kernel 6.19.0-rc5 if any VeraCrypt volume is mounted. This is true whether the VeraCrypt volume is partition-based or file-based. All previous kernels, including 6.17 are fine.

# Expected behavior
Laptop power light goes out
Disk activity light goes out
Laptop (Linux kernel) suspend normally

# Observed behavior
Power light remains on
Disk activity light remains on
Laptop (Linux kernel) fails to suspend

# Steps to reproduce
1. Mount any VeraCrypt volume normally 
2. Issue suspend command ('systemctl suspend') or suspend from the desktop environment

# Screenshots
N/A

# Your Environment

**VeraCrypt version:** 
1.26.24  (But also occurs using 1.26.14

**Hardware:**
Lenovo IdeaPad 5 15ITL05
11th Gen Intel® Core™ i5-1135G7 × 4
8.0 GiB Memory
Intel® Iris® Xe Graphics (TGL GT2)

**Operating Environment:**
Firmware ver: FHCN78WW
openSUSE Tumbleweed
64-bit
GNOME 49 Wayland
Linux 6.18.5-1-default

**Additional Notes**

I have opened the following bug reports, thinking that this is a kernel bug. However, they seem to believe that it is a a VeraCrypt bug. Assistance is needed in tracking down the problem in any case.

https://bugzilla.kernel.org/show_bug.cgi?id=220996

https://bugzilla.opensuse.org/show_bug.cgi?id=1256893

abhinavagarwal07:
Fixes #1660.

On kernel 6.18+, `systemctl suspend` hangs indefinitely when a VeraCrypt volume is mounted. Two independent changes combined to cause this:

1. **systemd v255** made `user.slice` freezing the default before suspend
2. **Kernel 6.18** (commit `d3906d8f3cee`) enabled `FUSE_SYNCFS` for all fuseblk mounts

When a VeraCrypt NTFS volume is mounted, ntfs-3g runs in `user.slice`. The suspend path freezes ntfs-3g, then the kernel sends `FUSE_SYNCFS` to it and blocks forever waiting for a response.

**Tested on:** Ubuntu 26.04 (kernel 7.0.0, systemd 259). Built with `NOGUI=1 WITHFUSE3=1`. Confirmed FUSE daemons land in `user.slice/user-1000.slice/session-*.scope`. Verified the sleep hook dismounts all volumes when invoked with `pre suspend`. Azure VMs lack ACPI S3, so the actual suspend hang could not be triggered end-to-end.

### DIFF
--- a/src/Build/CMakeLists.txt
+++ b/src/Build/CMakeLists.txt
@@ -213,6 +213,11 @@ install(DIRECTORY 	${VERACRYPT_BUILD_DIR}/usr/sbin
 install(DIRECTORY 	${VERACRYPT_BUILD_DIR}/usr/share
 		DESTINATION .
 		USE_SOURCE_PERMISSIONS)
+if(EXISTS "${VERACRYPT_BUILD_DIR}/lib/systemd/system-sleep")
+	install(DIRECTORY 	${VERACRYPT_BUILD_DIR}/lib/systemd/system-sleep
+			DESTINATION /lib/systemd
+			USE_SOURCE_PERMISSIONS)
+endif()
 set(CPACK_PACKAGING_INSTALL_PREFIX "/usr")
 
 # For packaging 
--- a/src/Main/GraphicUserInterface.cpp
+++ b/src/Main/GraphicUserInterface.cpp
@@ -1091,7 +1091,7 @@ namespace VeraCrypt
 
 			Connect (wxEVT_END_SESSION, wxCloseEventHandler (GraphicUserInterface::OnEndSession));
 #ifdef wxHAS_POWER_EVENTS
-			Gui->Connect (wxEVT_POWER_SUSPENDING, wxPowerEventHandler (GraphicUserInterface::OnPowerSuspending));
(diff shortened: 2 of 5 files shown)


idrassi:
The root-cause hypothesis is plausible but this implementation is not mergeable.

systemd documents that system-sleep hooks run while user.slice is frozen, so the hook can block when veracrypt --dismount tries to query VeraCrypt FUSE control file or unmount ntfs-3g. Also, --force does not pass umount -f on Linux in VeraCrypt current code path and the hook silently dismounts all users volumes on every suspend regardless of preferences.

Please split the wxGTK power-event change into a separate PR and investigate a delay-inhibitor or upstream systemd/kernel/ntfs-3g fix instead.

PR #1678, opened 2026-04-15
Fix arm64 macOS build with Homebrew

audriusbuika:
## Problem

On Apple Silicon, `./src/Build/build_veracrypt_macosx.sh -b` fails to build:
```bash
clang: error: unsupported option '-mssse3' for target 'arm64-apple-darwin24.6.0'
clang: error: unsupported option '-msse4.1' for target 'arm64-apple-darwin24.6.0'
clang: error: unsupported option '-msha'   for target 'arm64-apple-darwin24.6.0'
make[1]: *** [../Crypto/Sha2Intel.oshani] Error 1
```

Once the flag issue is worked around, the link step then fails with missing
arm64 symbols: `_aes_encrypt`, `_aes_hw_cpu_*`, and
`_sha256_compress_digest_armv8`.

## Root Cause

Two separate issues surface together on arm64:

1. The pattern rules in `src/Build/Include/Makefile.inc` (`%.oshani`,
   `%.osse41`, `%.ossse3`, `%.oaesni`, `%.oavx2`) always append x86-only
   flags such as `-mssse3 -msse4.1 -msha`. clang rejects those flags when
   targeting arm64, even though the source bodies are already guarded by
   `CRYPTOPP_BOOL_X86/X64` and would compile to empty objects.

2. `src/Volume/Volume.make` has no arm64 object list for local MacOSX
   builds. The build script sets `COMPILE_ASM=false` on arm64 hosts (yasm
   can't assemble x86 asm for an arm64-only build), which leaves the
   MacOSX branch empty, and the generic arm64 branch below it is never
   reached. As a result, no AES or SHA-256 backend ends up in the link.

## Fix

**`src/Build/Include/Makefile.inc`** — move the x86 feature flags into
variables that are empty when `CPU_ARCH=arm64`. x86/x64 builds keep the
same flags they had before; arm64 compiles the same sources without
them, and the existing `#ifdef` guards take care of the rest.

**`src/Volume/Volume.make`** —
1. Add an `else ifeq "$(CPU_ARCH)" "arm64"` branch inside the MacOSX
   block that pulls in `Aes_hw_armv8.oarmv8crypto`, `Aescrypt.o`, and
   `sha256_armv8.oarmv8crypto` — the symbols the linker was missing.
2. On arm64, compile `Sha2Intel`, `blake2s_SSE41`, `blake2s_SSSE3`, and
   `opt_avx2` as plain `.o` files so they skip the x86-flag pattern rules.

## Verification

The build now completes on arm64 (macOS 24.6, SDK 26.2) and produces a
working `src/Main/VeraCrypt` binary. x86_64 and universal-binary release
builds are untouched — the new paths only activate when `CPU_ARCH=arm64`.


idrassi:
Thanks for investigating the Apple Silicon/Homebrew build failure.

The local arm64 build issue is real, but I do not think this PR cannot be merged as-is because the `CPU_ARCH=arm64` conditions also affect the normal non-local macOS release build on Apple Silicon.

On Apple Silicon, `CPU_ARCH` is derived from `uname -m`, so it is `arm64` even when the build is not local-only. In that non-local path, `src/Makefile` still builds a universal `x86_64 + arm64` binary. With this PR, `src/Volume/Volume.make` therefore compiles `blake2s_SSE41`, `blake2s_SSSE3`, `Sha2Intel` and `Argon2/src/opt_avx2` as plain `.o` files for the universal build instead of using the feature-specific `.osse41`, `.ossse3`, `.oshani`, and `.oavx2` rules.

The `opt_avx2` case is a functional regression: without `-mavx2`, `__AVX2__` is not defined, so `fill_segment_avx2` compiles to the stub returning `ARGON2_INCORRECT_PARAMETER`. The x86 runtime dispatch calls this function when `HasSAVX2` is true, so the x86_64 slice of a universal macOS build produced on Apple Silicon can fail on AVX2-capable Intel machines.

Please scope the plain-object fallback to the local arm64 build only, for example by keying it on `LOCAL_DEVELOPMENT_BUILD=true` and `CPU_ARCH=arm64` rather than on `CPU_ARCH=arm64` alone. The global `Makefile.inc` flag stripping should either be avoided or constrained by the same condition.

src/Volume/Volume.make
```diff
@@ -95,6 +98,14 @@ else
...
 endif
 
+ifeq "$(CPU_ARCH)" "arm64"
+# x86-only intrinsics sources are compiled as plain objects on arm64
+# (their bodies are #ifdef-gated to x86/x64 and become empty translation units)
+OBJS += ../Crypto/blake2s_SSE41.o
+OBJS += ../Crypto/blake2s_SSSE3.o
+OBJS += ../Crypto/Sha2Intel.o
+OBJS += ../Crypto/Argon2/src/opt_avx2.o
+else
```
idrassi, line 108: This condition is too broad. `CPU_ARCH=arm64` is also true for the normal non-local macOS build on Apple Silicon, which still produces a universal `x86_64 + arm64` binary. In that path this block makes the x86_64 slice compile `Argon2/src/opt_avx2.c` as plain `.o` instead of `.oavx2`, so `__AVX2__` is not defined and `fill_segment_avx2()` becomes the stub that returns `ARGON2_INCORRECT_PARAMETER`. Please scope this fallback to the local arm64-only case, e.g. `LOCAL_DEVELOPMENT_BUILD=true` plus `CPU_ARCH=arm64`, not `CPU_ARCH=arm64` alone.

idrassi (changes requested):

src/Build/Include/Makefile.inc
```diff
@@ -16,33 +16,49 @@ clean:
 	@echo Cleaning $(NAME)
 	rm -f $(APPNAME) $(NAME).a $(OBJS) $(OBJSEX) $(OBJSNOOPT) $(OBJSHANI) $(OBJAESNI) $(OBJSSSE41) $(OBJSSSSE3) $(OBJSAVX2) $(OBJARMV8CRYPTO) $(OBJS:.o=.d) $(OBJSEX:.oo=.d) $(OBJSNOOPT:.o0=.d) $(OBJSHANI:.oshani=.d) $(OBJAESNI:.oaesni=.d) $(OBJSSSE41:.osse41=.d) $(OBJSSSSE3:.ossse3=.d) $(OBJSAVX2:.oavx2=.d) $(OBJARMV8CRYPTO:.oarmv8crypto=.d) *.gch
 
+# On arm64 hosts, x86 feature flags are rejected by clang.
+# Source files are guarded by CRYPTOPP_BOOL_X86/X64 so they produce empty objects on arm64.
+ifeq "$(CPU_ARCH)" "arm64"
```
idrassi, line 21: This has the same scoping issue as the `Volume.make` change: `CPU_ARCH=arm64` is true on Apple Silicon even for the normal non-local macOS build, which still targets a universal `x86_64 + arm64` binary.
For that universal path, the x86_64 slice still needs these feature flags. In particular, clearing `X86_SHANI_FLAGS` makes `Sha2Intel.oshani` compile without `-msha`: with Apple clang this fails because `_mm_sha256msg1_epu32` requires the `sha` target feature.
Please either leave these pattern rules unchanged and avoid selecting the x86-specific suffix objects for local arm64-only builds, or gate this flag clearing on the local arm64-only case, e.g. `LOCAL_DEVELOPMENT_BUILD=true` plus `CPU_ARCH=arm64`, not `CPU_ARCH=arm64` alone.

idrassi (changes requested):

damianrickard:
Confirming this still reproduces on current `master` (d26216c) — macOS 26.5 (Tahoe), Xcode 21 / clang, arm64 host, building with `./src/Build/build_veracrypt_macosx.sh -b` against Homebrew wxWidgets 3.3.2:

```
clang: error: unsupported option '-msha' for target 'arm64-apple-darwin25.5.0'
make[1]: *** [../Crypto/Sha2Intel.oshani] Error 1
```

One observation that may help reviewers: on current `master` the `Volume.make` arm64 local-dev branch (`ifeq "$(LOCAL_DEVELOPMENT_BUILD):$(CPU_ARCH)" "true:arm64"`) already pulls in the ARMv8 AES/SHA objects, so I hit **no** missing-symbol link errors — for me the only remaining failure was the `Makefile.inc` pattern rules (`%.oshani`/`%.osse41`/`%.ossse3`/`%.oavx2`) unconditionally appending `-mssse3 -msse4.1 -msha`. Your `Makefile.inc` change is the essential fix; the `Volume.make` portion may now partially overlap with master. After applying the patch I get a working arm64 `src/Main/VeraCrypt`. Would be great to see this merged. 🙏

idrassi:
Thanks for the feedback.

I can't merge this PR as-is because it gates on CPU_ARCH=arm64, which also applies to normal non-local macOS builds on Apple Silicon. Those builds still produce universal binaries, so the x86_64 slice must keep the feature-specific x86 flags and object rules.

The ARM AES/SHA missing symbol part has already been fixed on master, so the remaining fix should be narrower: only avoid the x86 feature-suffix objects for LOCAL_DEVELOPMENT_BUILD=true + CPU_ARCH=arm64.

Since this PR is now stale and conflicts with master, I'll close it and land a scoped commit, with credit for the original report and confirmation.

idrassi:
I have push the commit https://github.com/veracrypt/VeraCrypt/commit/d33b23a9be2e61bd8f3a3c3056a09ca7619fd8af with proper fixes.

Issue #1446, opened 2024-11-23
freebsd bug veracrypt not able to create a basic file container volume


justanotherscript:
# Expected behavior
Changes between 1.26.7 and1.26.14 not all but to describe the error (25 August 2024) :
FreeBSD:
    Fix privilege escalation prompts not showing up (GH #1349)
    Support automatic detection and mounting of ext2/3/4, exFAT, NTFS filesystems (GH #1350)
    Use correct Disk Utility location when "check filesystem" is ran (GH #1273)

# Observed behavior
privilege escalation prompts are appearing for example when trying to create a a basic file container volume but after typing in the correct sudo password "Not enough data available" appears as error.

# Steps to reproduce
creating a new veracrypt basic/ standard file container volume

# Screenshots

# Your Environment
![1](https://github.com/user-attachments/assets/aba09120-4f3a-4c36-acc7-815a14feda9e)
![2](https://github.com/user-attachments/assets/43f7e3e7-01f6-4e7f-a927-3cc5667e38ea)
![3](https://github.com/user-attachments/assets/377adc1b-8056-4ecf-839a-5c93da19439d)
![4](https://github.com/user-attachments/assets/ca10f2cc-249c-4d9d-a46b-8fb2c4e44b70)
![5](https://github.com/user-attachments/assets/3f99cbe6-e3fa-40cb-85d1-b1d71b800ae7)
![6](https://github.com/user-attachments/assets/dbe39b98-5c56-4fc2-9b47-6329849bcbe5)
![7](https://github.com/user-attachments/assets/6bc0dab5-43d4-4a49-ac3e-5b950afc53ad)
![8](https://github.com/user-attachments/assets/2559ecb3-3b0c-4cc1-b362-a6cbb03f2a80)
![9](https://github.com/user-attachments/assets/05760e78-74a1-4984-bb46-96a86be1f007)
![10](https://github.com/user-attachments/assets/4c0e550c-69aa-406d-9448-2c1515f77ad3)
![11](https://github.com/user-attachments/assets/520cff16-900c-464c-82bb-80e16eb29735)
![12](https://github.com/user-attachments/assets/fa2db28e-21b2-4fd5-8e23-1727b1429720)
![13](https://github.com/user-attachments/assets/4f41e7f8-7d6c-4d44-a578-7400229b2883)
![14_sudo_promt](https://github.com/user-attachments/assets/c34893c9-8353-4730-80f5-14e9c3bc2803)
[...]

Issue #844, opened 2021-10-26
Bad File Descriptor VeraCrypt::CoreService::StartElevated:567 (Bug)


ghost:
I'm running Ubuntu 21.10. I've had the same issue on 21.04 as well. I am just about tired of having the issue so I'm posting about it. Basically, trying to mount an encrypted image, enter password, followed by entering admin password gets the error "**Bad File Descriptor VeraCrypt::CoreService::StartElevated:567**". But if you try to mount it again it succeeds. This happens with creating containers as well as mounting them. I've screen recorded it to show what I'm on about. I've tested this with EXT3, EXT4 and NTFS. All the same. I am the only account on the system and have SUDO permissions. 

I've tried changing the exec= command in the .desktop file to "sudo /usr/bin/veracrypt" but it changes nothing, and using "pkexec /usr/bin/veracrypt" does requests my fingerprint or password but then displays nothing.

https://youtu.be/5Zt-M1LnwyI

jypgithub:
I have the same problem with Ubuntu 18.04.

idrassi:
I was not able to reproduce this issue on my various Ubuntu VMs. So there must be some configuration that triggers this behavior and this configuration doesn't look to be widespread.

The error message give some clues(it happens in line 567 of [CoreService.cpp](https://github.com/veracrypt/VeraCrypt/blob/master/src/Core/Unix/CoreService.cpp#L567)):
`throw_sys_if (fcntl (outPipe->GetReadFD(), F_SETFL, 0) == -1);`

"Bad File Descriptor" means that `outPipe` became invalid and the only possibility is that it was closed at [line 506](https://github.com/veracrypt/VeraCrypt/blob/master/src/Core/Unix/CoreService.cpp#L506) because the forked sudo process is not responding after 6 seconds which triggers `TimeOut` exception.

It is strange that the sudo process doesn't exit or respond within 6 seconds. The only possibility I see is that the forked sudo process starts "too late" after we write the admin password to the pipe at [line 461](https://github.com/veracrypt/VeraCrypt/blob/master/src/Core/Unix/CoreService.cpp#L461).

One idea is too add a 1 second sleep before writing admin password to forked process pipe.This may help.

Meanwhile, could you try starting VeraCrypt with the switch "use-dummy-sudo-password"? This changes the way sudo forking is handled and it's possible that it will help.

idrassi:
I have implement the 1 second sleep idea explain above.
[...]

PR #1687, opened 2026-04-23
Fix undefined behavior in StartElevated stderr read loop (#1550)

### LINKED ISSUE #1550: Not enough data available
# Expected behavior

1. Veracrypt should provide useful error messages
2. optionally run without sudo, asking for the root password for privileged actions like formatting

# Observed behavior

When Veracrypt is run on Arch Linux by a user without sudo permissions, an error is shown during the format process with the message: "Not enought data available"
When the Format button is used again, the error mesage changes to: "VeraCrypt::File::Write:395"

Using Veracrypt on Debian (without sudo available) results in a proper error message about missing permissions. (Without sudo, Veracrypt needs to run entirely as root, which is an issue on its own)

# Steps to reproduce

1. Setup a new container with ext4 on Arch Linux
2. Click format
3. Dont be in the sudoers file
4. Find weird error message

# Screenshots

Image

Image

# Your Environment

**VeraCrypt version:** 
1.26.24 
(also tested with 1.26.20)

**Operating system and version:** 
Arch Linux

**System type:** 
64-bit

curious-rabbit:
The read loop that captures stderr from the sudo child process used `vector<char> buffer(4096); buffer.clear();` followed by `read(fd, &buffer[0], buffer.capacity())`. This has two instances of undefined behavior:

1. `operator[](0)` on a vector with `size() == 0` violates the C++ standard precondition `n < size()`. libstdc++ built with `-D_GLIBCXX_ASSERTIONS` aborts the process with:

     stl_vector.h:1128: Assertion '__n < this->size()' failed.

2. `buffer.begin() + bytesRead` on the same empty vector constructs an iterator past `end()`, also UB.

`-D_GLIBCXX_ASSERTIONS` is in the default build flags of Arch Linux, Fedora, and several other distributions. On those systems, the unprivileged helper process aborts as soon as sudo writes anything to stderr (a password prompt, a 'user is not in the sudoers file' error, etc.). The main process then sees EOF on the service output pipe, and throws `InsufficientData`, which renders to the user as 'Not enough data available'. A second mount attempt fails at `File::Write` because the helper is dead and the pipe is broken, producing the bare message 'VeraCrypt::File::Write:395'.

Fix by replacing `buffer` with a plain `char[4096]` and using `reserve(4096)` on `errOutput` to preserve the original pre-allocation intent. No behavioral change on systems where the UB happened to work; aborts are eliminated on systems where the assertions fire.

Reported-by: multiple users
fixes #1550 #1446 #844

### DIFF
--- a/src/Core/Unix/CoreService.cpp
+++ b/src/Core/Unix/CoreService.cpp
@@ -503,9 +503,9 @@ namespace VeraCrypt
 		throw_sys_if (fcntl (outPipe->GetReadFD(), F_SETFL, O_NONBLOCK) == -1);
 		throw_sys_if (fcntl (errPipe.GetReadFD(), F_SETFL, O_NONBLOCK) == -1);
 
-		vector <char> buffer (4096), errOutput (4096);
-		buffer.clear ();
-		errOutput.clear ();
+		char buffer[4096];
+		vector <char> errOutput;
+		errOutput.reserve (4096);
 
 		Poller poller (outPipe->GetReadFD(), errPipe.GetReadFD());
 		int status, waitRes;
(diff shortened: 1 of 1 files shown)


idrassi:
Good catch!
Thank you for fixing this illusive bug. 
I'm merging it.

PR #1695, opened 2026-04-25
Linux: allow mounting NTFS volumes with ntfs3

mammothcoding:
## Summary

This PR allows Linux users to mount NTFS volumes with the in-kernel `ntfs3` driver instead of `ntfs-3g`/`fuseblk`.

It intentionally does not add sleep hooks, automatic dismounting, or arbitrary filesystem type passthrough. It adds a narrow Linux-only `ntfs3` path for NTFS volumes.

## Motivation

This is related to #1660.

On recent Linux/systemd combinations, NTFS VeraCrypt volumes mounted through `ntfs-3g` appear as `fuseblk`. During suspend/hibernate, systemd may freeze `user.slice` while the kernel sleep path tries to sync FUSE-backed filesystems. If the userspace `ntfs-3g` daemon is frozen, the sleep path can hang.

Mounting the NTFS filesystem with the Linux in-kernel `ntfs3` driver avoids the FUSE userspace daemon path.

## What changed

- Accept `ntfs3` as a valid Linux value for `--filesystem`.
- Canonicalize the explicit mount filesystem type to lowercase `ntfs3`.
- Keep the existing `VolumeCreationOptions::FilesystemType::NTFS` mapping, because `ntfs3` is a Linux mount driver for the NTFS filesystem, not a separate on-disk format.
- Add a Linux preference named "Mount NTFS volumes with the Linux kernel ntfs3 driver".
- When that preference is enabled, detect NTFS filesystems with `blkid` and mount them with `ntfs3` automatically. This applies to the Linux native dm-crypt path and to the fallback auxiliary-image mount path.
- Make the preference apply to favorite volumes, because favorites use the default mount options.
- Document in command-line help, GUI preference text, and the HTML documentation that the `ntfs3` kernel module must be available and allowed by the distribution. Some distributions blacklist it by default and require administrator action before it can be loaded.
- Add a Linux help example for mounting an NTFS volume using `--filesystem=ntfs3`.

## Local validation

I reproduced the affected setup on openSUSE Tumbleweed with:

- Kernel: `6.19.11-1-default`
- systemd: `259.5+suse`
- VeraCrypt: `1.26.24`
- `ntfs-3g`: `2022.10.3`
- `fuse3`: `3.18.2`

The affected NTFS VeraCrypt volume previously mounted as `fuseblk` and hibernation could hang. I then opened the VeraCrypt volume without mounting its filesystem and mounted the virtual device with:

```bash
mount -t ntfs3 -o uid=1000,gid=1000,umask=077 /dev/mapper/veracrypt1 /home/myuser/SecVol
```

`findmnt` then reported:

```text
/home/myuser/SecVol /dev/mapper/veracrypt1 ntfs3
```

With the VeraCrypt volume left mounted as `ntfs3`, hibernation and resume completed successfully and the volume remained mounted after resume.

Patched binary validation:

- Built successfully on the same openSUSE Tumbleweed system with `make NOGUI=1`.
- Built successfully on the same system with the default GUI build (`make`).
- `veracrypt --text --help` includes the new `ntfs3` help text and example.
- `--filesystem=ntfs3` is accepted by the command-line parser.
- An arbitrary filesystem value such as `--filesystem=definitely-not-allowed` is still rejected as an unknown option.

Live patched-binary mount validation:

- Mounted the same NTFS VeraCrypt volume through the patched binary using `--filesystem=ntfs3`.
- `findmnt` reported the mounted filesystem as `ntfs3` on `/dev/mapper/veracrypt1`.
- The volume remained visible to VeraCrypt in its normal slot after the filesystem mount.

Favorite-volume validation:

- Enabled the new `MountNtfsWithNtfs3` preference.
- Mounted the same NTFS VeraCrypt volume through `--load-preferences --auto-mount favorites`; `findmnt` reported `ntfs3`.
- Mounted the same favorite volume through the GUI menu `Favorites -> Mount Favorite Volumes`; `findmnt` reported `ntfs3`.
- With the GUI-mounted favorite volume left mounted as `ntfs3`, hibernation and resume completed successfully and the volume remained mounted after resume.


idrassi:
@mammothcoding 
Thanks for the detailed testing and for keeping the PR narrow.

Your results are useful confirmation for #1660 conclusion.

I agree with the general direction of this P  but I can't merge it as-is. Please revise the following:

1. Gate the new GUI preference strictly to Linux. The persistence and mount behavior are Linux-only but the checkbox and 
help text are currently added in the shared wx UI path. Non-Linux Unix builds should not expose a Linux-only ntfs3 option.

2. Avoid duplicating `DetectFilesystemType` in both CoreUnix.cpp and CoreLinux.cpp. Please consolidate it into one Linux helper or otherwise avoid duplicated logic.

3. Make the NTFS detection use uncached probing, e.g. `blkid -p -o value -s TYPE -- <device>`, because VeraCrypt
reuses /dev/mapper/veracryptN device names. 

4. Clarify the documentation/help text around detection and fallback. The preference should be described as applying only when VeraCrypt detects NTFS via `blkid`. If detection fails or if ntfs3 is unavailable, users should not be left with the impression that the FUSE path was definitely avoided.

5. Keep this opt-in and disabled by default. Linux NTFS mounts can use different backends depending on kernel, installed packages and distribution policy (ntfs-3g/fuseblk, lowntfs-3g, ntfs3 and possibly newer kernel ntfs implementations). VeraCrypt shouldn't override the default NTFS backend for existing users unless they explicitly opt in.

Once these points are addressed, this can be considered as an opt-in mitigation for affected Linux users.

src/Main/Forms/Forms.cpp
```diff
@@ -1947,6 +1947,14 @@ PreferencesDialogBase::PreferencesDialogBase( wxWindow* parent, wxWindowID id, c
...
 	FilesystemSizer->Add( fgSizer3, 1, wxEXPAND, 5 );
 
+	MountNtfsWithNtfs3CheckBox = new wxCheckBox( FilesystemSizer->GetStaticBox(), wxID_ANY, _("LINUX_PREF_MOUNT_NTFS_WITH_NTFS3"), wxDefaultPosition, wxDefaultSize, 0 );
+	FilesystemSizer->Add( MountNtfsWithNtfs3CheckBox, 0, wxALL, 5 );
+
+	wxStaticText* MountNtfsWithNtfs3HelpStaticText;
+	MountNtfsWithNtfs3HelpStaticText = new wxStaticText( FilesystemSizer->GetStaticBox(), wxID_ANY, _("LINUX_PREF_MOUNT_NTFS_WITH_NTFS3_HELP"), wxDefaultPosition, wxDefaultSize, 0 );
+	MountNtfsWithNtfs3HelpStaticText->Wrap( 600 );
+	FilesystemSizer->Add( MountNtfsWithNtfs3HelpStaticText, 0, wxLEFT|wxRIGHT|wxBOTTOM|wxEXPAND, 5 );
+
```
idrassi: This is currently added in the shared generated wx UI path. Since the preference is Linux-only, please ensure this checkbox and its help text are only exposed on `TC_LINUX` builds, or are explicitly hidden on non-Linux Unix platforms.

src/Core/Unix/Linux/CoreLinux.cpp
```diff
@@ -30,6 +30,29 @@
...
+
+			try
+			{
+				return StringConverter::ToLower (StringConverter::Trim (Process::Execute ("blkid", args, 2000)));
+			}
+			catch (...)
+			{
+				return string();
+			}
+		}
```
idrassi: This duplicates the `DetectFilesystemType` helper added in `CoreUnix.cpp`. Please consolidate this into a single Linux helper used by both the native dm-crypt path and the auxiliary-image path.

src/Core/Unix/CoreUnix.cpp
```diff
@@ -566,6 +567,28 @@ namespace VeraCrypt
 		return GetMountedFilesystems (DevicePath(), mountPoint).size() == 0;
 	}
 
+#ifdef TC_LINUX
+	static string DetectFilesystemType (const DevicePath &devicePath)
+	{
+		list <string> args;
+		args.push_back ("-o");
```
idrassi, line 574: Please use uncached probing here, e.g. add `-p` so this calls `blkid -p -o value -s TYPE -- <device>`. VeraCrypt reuses `/dev/mapper/veracryptN` names, so relying on cached blkid results would be a bad failure mode.

src/Core/Unix/Linux/CoreLinux.cpp
```diff
@@ -30,6 +30,29 @@
 
 namespace VeraCrypt
 {
+	namespace
+	{
+		string DetectFilesystemType (const DevicePath &devicePath)
+		{
+			list <string> args;
+			args.push_back ("-o");
```
idrassi: Please use uncached probing here, e.g. add `-p` so this calls `blkid -p -o value -s TYPE -- <device>`. VeraCrypt reuses `/dev/mapper/veracryptN` names, so relying on cached blkid results would be a bad failure mode.

src/Common/Language.xml
```diff
@@ -1473,6 +1473,8 @@
     <entry lang="en" key="LINUX_PREF_KERNEL_CRYPT">Do not use kernel cryptographic services</entry>
     <entry lang="en" key="LINUX_PREF_TAB_MOUNT_OPTIONS_FS">Filesystem</entry>
     <entry lang="en" key="IDT_LINUX_PREF_TAB_MOUNT_OPTIONS">Mount options:</entry>
+    <entry lang="en" key="LINUX_PREF_MOUNT_NTFS_WITH_NTFS3">Mount NTFS volumes with the Linux kernel ntfs3 driver</entry>
+    <entry lang="en" key="LINUX_PREF_MOUNT_NTFS_WITH_NTFS3_HELP">When enabled, VeraCrypt detects NTFS filesystems and mounts them with ntfs3 instead of ntfs-3g/fuseblk. This can avoid suspend or hibernate hangs on systems where user-space FUSE filesystems are frozen while the kernel is syncing filesystems. The ntfs3 kernel module must be available and allowed by the distribution.</entry>
```
idrassi: Please make this text more explicit that the preference applies only when VeraCrypt detects `TYPE=ntfs` via `blkid`. If detection fails, the normal mount path is used. If ntfs3 is unavailable or blacklisted by the distribution, users should not be left with the impression that the FUSE path was definitely avoided.

Translations/Language.ru.xml
```diff
@@ -1473,6 +1473,8 @@
     <entry lang="ru" key="LINUX_PREF_KERNEL_CRYPT">Не использовать криптографические службы ядра</entry>
     <entry lang="ru" key="LINUX_PREF_TAB_MOUNT_OPTIONS_FS">Файловая система</entry>
     <entry lang="ru" key="IDT_LINUX_PREF_TAB_MOUNT_OPTIONS">Опции монтирования:</entry>
+    <entry lang="ru" key="LINUX_PREF_MOUNT_NTFS_WITH_NTFS3">Монтировать NTFS-тома через драйвер ядра Linux ntfs3</entry>
+    <entry lang="ru" key="LINUX_PREF_MOUNT_NTFS_WITH_NTFS3_HELP">Если включено, VeraCrypt определяет файловые системы NTFS и монтирует их через ntfs3 вместо ntfs-3g/fuseblk. Это может предотвратить зависания при ждущем режиме или гибернации на системах, где пользовательские FUSE-файловые системы замораживаются, пока ядро синхронизирует файловые системы. Модуль ядра ntfs3 должен быть доступен и разрешён дистрибутивом.</entry>
```
idrassi: It is preferable to keep entries in English for this PR and have a separate PR for translation so that it can reviewed by maintainer of the Russian translation.

idrassi (changes requested):

mammothcoding:
Thanks for the note about translation changes. I removed the Russian language and Russian documentation updates from this PR in 2411942, so this PR now keeps only the code changes plus English strings/docs. If this PR is accepted, I can follow up with a separate Russian translation PR based on the final English text.

idrassi:
@mammothcoding Thanks for the changes. I did a small update to add new entries to XML language files to make Validate XML pass

PR #1699, opened 2026-04-27
Some minor fixes to Norwegian Bokmål translation

sandakersmann:
Some minor fixes to Norwegian Bokmål translation


idrassi:
Thanks

PR #1705, opened 2026-04-30
Increment version to 1.26.28 in Language.nl.xml

sandakersmann:
Increment version to 1.26.28 in Language.nl.xml as well.


idrassi:
Thank you for catching this...my find-replace all failed me!

PR #1707, opened 2026-05-01
Fix whitespace issues in translation files

sandakersmann:
Fix whitespace issues in translation files.

Also fixed prog-version in /src/Common/Language.xml


idrassi:
Thanks

PR #1725, opened 2026-05-11
Update Polish translation

Matthaiks:


idrassi:
@Matthaiks 
I have one small request before merging: in `MACOSX_APFS_SYNTHESIZED_DEVICE`, please move `{1}` after `magazynu APFS`. This placeholder is an optional hint such as ` (/dev/rdisk...)`, so the current text can render awkwardly as `partycję (/dev/rdisk...) magazynu APFS` or leave a double space when the hint is empty.

For example:

  `Zamiast tego wybierz fizyczną partycję magazynu APFS{1}.`

Matthaiks:
@idrassi 
Done.

idrassi:
Thank you for the quick feedback.

PR #1727, opened 2026-05-12
Update Language.de.xml

Mr-Update:
- Translation completed


idrassi:
@Mr-Update 
Before merging, I have comments.

In MACOSX_APFS_SYNTHESIZED_DEVICE, the placeholder {1} is already generated by the code with a leading space when a recommended physical store path is available, so the German string shouldn't add another space before it:

      `Wählen Sie stattdessen die physische APFS-Speicherpartition{1} aus.`

Also, for MACOSX_APFS_SYSTEM_STORE:  "APFS-Speicherort" may be misleading because Apple/diskutil uses "APFS Physical Store" for the storage object backing an APFS Container, not for a general location/path: https://keith.github.io/xcode-man-pages/diskutil.8.html

Apple German Disk Utility documentation also uses APFS terms such as Container, Volume, and APFS-Partition in this context: https://support.apple.com/de-de/guide/disk-utility/dskutl1005/mac

So it looks like "physische APFS-Speicherpartition" is better in this context.

Mr-Update:
@idrassi I've made the changes.

idrassi:
Thank you!

PR #1732, opened 2026-05-15
Update Language.zh-cn.xml

nkh0472:


idrassi:
@nkh0472 Thanks. I noticed two small cleanup points before merging:

1. Please remove the extra blank line added before `</localization>`.
2. For `MACOSX_APFS_SYNTHESIZED_DEVICE`, `{1}` is already generated with a leading space and parentheses when a recommended device path is available, so the sentence should avoid adding another space before `{1}` and should also read naturally when `{1}` is empty (the English version is made in such a way).

I checked a better wording with AI and it proposed:

```xml
<entry lang="zh-cn" key="MACOSX_APFS_SYNTHESIZED_DEVICE">所选设备 '{0}' 是 APFS 合成容器或卷，不能作为原始 VeraCrypt 卷的承载设备使用。\n\n请改选物理 APFS 存储分区{1}。</entry>
```

Sorry for this tricky `{1}`, other translators had the same issue.

nkh0472:
> 1. Please remove the extra blank line added before `</localization>`.
> 2. For `MACOSX_APFS_SYNTHESIZED_DEVICE`, `{1}` is already generated with a leading space and parentheses when a recommended device path is available, so the sentence should avoid adding another space before `{1}` and should also read naturally when `{1}` is empty (the English version is made in such a way).

solved

PR #1736, opened 2026-05-18
Update FUSE package version for Debian and Ubuntu

thurask:
Missed this in testing because I had the old package left over on my Debian system, but libfuse3-3 was dropped in Debian 13/Ubuntu 25.10 and replaced with libfuse3-4; update the CMakeLists.txt to account for this


idrassi:
Thanks!

PR #1737, opened 2026-05-18
Update Polish translation

Matthaiks:


idrassi:
@Matthaiks 
Before merging, could you please adjust the technical wording in `LINUX_PREF_MOUNT_NTFS_WITH_KERNEL_DRIVER_HELP` so that the Linux driver identifiers remain lowercase: `ntfs` and `ntfs3`? They refer to the exact Linux filesystem/driver names used by the mount logic, and not just translated NTFS names here. It would also be clearer to say that `Linux 7.1 or later` refers to the Linux kernel version.

For example, AI proposes the following sentence:

`VeraCrypt używa ntfs, gdy zostanie on jednoznacznie rozpoznany jako nowoczesny sterownik odczytu/zapisu albo jest oczekiwany w jądrze Linuksa 7.1 lub nowszym, a w przeciwnym razie używa ntfs3.`

After that adjustment, this PR is good to merge.

Matthaiks:
@idrassi 
Made the adjustment.

idrassi:
@Matthaiks Thanks for the quick feedback.

PR #1738, opened 2026-05-18
Update to Norwegian Bokmål translation

sandakersmann:
Update to Norwegian Bokmål translation


idrassi:
@sandakersmann 
Before merging, could you please confirm one wording detail? For `LINUX_ADMIN_PW_QUERY`, the English source says "Enter your user password or administrator password", while the new Norwegian text only mentions the administrator password. Since this dialog is used for sudo/elevation and may require the current user password, should the Norwegian translation preserve both meanings?

Also please confirm that `NTFS-bakgrunn` is the intended Norwegian wording for “NTFS backend” in `LINUX_KERNEL_NTFS_DRIVER_UNAVAILABLE`.

sandakersmann:
You are correct. Fixed :)

idrassi:
Thanks!

PR #1748, opened 2026-05-28
Update Language.zh-cn.xml

nkh0472:
Translate NTFS mount options to Chinese


idrassi:
Thank you. Before merging, could you please address and share your feedback on the points below raised by the AI review system:

- Please avoid "内核内置". It can imply a driver built directly into the kernel image, while "in-kernel" here also covers loadable kernel modules. "Linux 内核驱动程序" or "Linux 内核 NTFS 驱动程序" would be more accurate.
- In the help text, "当 NTFS 被明确识别为现代读写驱动程序" should refer to the literal mount/filesystem selector `ntfs`, not NTFS as the filesystem. For example: "当 ntfs 明确对应现代读写驱动程序，或预计在 Linux 内核 7.1 或更高版本中可用时，VeraCrypt 使用 ntfs；否则使用 ntfs3。"
- Please also use "Linux 内核 7.1" instead of "Linux 7.1" to avoid ambiguity.

Once these wording points are addressed, this should be good to merge.

nkh0472:
> Thank you. Before merging, could you please address and share your feedback on the points below raised by the AI review system:
> 
> * Please avoid "内核内置". It can imply a driver built directly into the kernel image, while "in-kernel" here also covers loadable kernel modules. "Linux 内核驱动程序" or "Linux 内核 NTFS 驱动程序" would be more accurate.
> * In the help text, "当 NTFS 被明确识别为现代读写驱动程序" should refer to the literal mount/filesystem selector `ntfs`, not NTFS as the filesystem. For example: "当 ntfs 明确对应现代读写驱动程序，或预计在 Linux 内核 7.1 或更高版本中可用时，VeraCrypt 使用 ntfs；否则使用 ntfs3。"
> * Please also use "Linux 内核 7.1" instead of "Linux 7.1" to avoid ambiguity.

Thanks for the suggestions. Adopted all of them.

idrassi:
Thanks.

PR #1756, opened 2026-06-10
Honor --no-size-check in CLI volume creation

damianrickard:
## Problem

The documented `--no-size-check` switch ("Disable check of container size against disk free space") has **no effect** when creating a file-hosted container via `--text --create`. Creating a container larger than current free space fails with `Error: Incorrect volume size`, even though that's a valid request on sparse-capable filesystems (APFS, ext4, NTFS) and is exactly what the flag exists to permit.

## Root cause

In `TextUserInterface.cpp`, the volume-creation path clamps `maxVolumeSize` to available free disk space and never consults `CmdLine->ArgDisableFileSizeCheck`. The flag *is* honored by the GUI wizard (`Forms/VolumeSizeWizardPage.cpp`) but was never wired into the text UI.

## Fix

Skip the free-space clamp when `--no-size-check` is set, mirroring the GUI behavior. One-line guard; no change to default behavior.

## Verification

Built on macOS arm64; `--text --create --size 3500G --quick --no-size-check ...` now succeeds, producing a sparse 3.42 TiB container (256 KB physical on APFS). Without the flag, the free-space check still applies as before.

### DIFF
--- a/src/Main/TextUserInterface.cpp
+++ b/src/Main/TextUserInterface.cpp
@@ -766,7 +766,10 @@ namespace VeraCrypt
 				if (options->Type == VolumeType::Normal && wxDirExists(parentDir) && wxGetDiskSpace (parentDir, nullptr, &diskSpace))
 				{
 					AvailableDiskSpace = (uint64) diskSpace.GetValue ();
-					if (maxVolumeSize > AvailableDiskSpace)
+					// Honor --no-size-check so a (sparse) file container larger than the
+					// current free space can be created from the command line, mirroring
+					// the GUI wizard behavior (see VolumeSizeWizardPage.cpp).
+					if (maxVolumeSize > AvailableDiskSpace && !CmdLine->ArgDisableFileSizeCheck)
 						maxVolumeSize = AvailableDiskSpace;
 				}
 			}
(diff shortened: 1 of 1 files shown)


idrassi:
Thanks for the sharing the fix for this issue. That being, I can't merge this PR in its current state because of the following:

- The commit is marked as co-authored by Claude Fable 5: While I'm not against the use of AI for development, commits should be authored by humans only to take full responsibility and ownership of the changes. co-ownership with AI is not accepted. Please amend the commit to remove the Claude Fable 5 co-author.
- When --no-size-check is set, maxVolumeSize stays at TC_MAX_VOLUME_SIZE_GENERAL instead of being clamped to AvailableDiskSpace. That's correct for an explicit --size value, but the --size max sentinel at lines 774–784 (and the interactive max choice at lines 808–812) also rely on maxVolumeSize.  With your current change, --size max --no-size-check would resolve to TC_MAX_VOLUME_SIZE_GENERAL (≈9.2 EB), which is clearly not intended. The max sentinel should still resolve to the smaller of the theoretical maximum and the actual free space.

Could you update both max paths to use:

```cpp
  options->Size = min (maxVolumeSize, AvailableDiskSpace);
```

This preserves current behavior when --no-size-check is absent (maxVolumeSize is already clamped) and uses AvailableDiskSpace when the check is disabled. Please also update the stale comment that says maxVolumeSize is guaranteed to be ≤ AvailableDiskSpace.

Once the commit is amended to include these changes and remove Claude Fable 5 co-author (if you take full responsability and ownership of the changes), I will merge this PR.

idrassi:
I updated the max paths on my side.

PR #1769, opened 2026-06-14
Translate English entries to Chinese in XML

nkh0472:


idrassi:
Thanks. As usual, I ran an AI review and it shared the feedback below. Can you please check it?

One small wording issue needs adjustment before merge: in `PIM_RESET_ON_KDF_CHANGE_CONFIRM`, the text currently refers to “新建” and “使用 PIM”, but the existing Simplified Chinese UI labels in this dialog are “新密码” (`IDT_NEW`) and “调整 PIM” (`IDC_NEW_PIM_ENABLE`). To avoid pointing users to labels that do not appear in the localized dialog, please update this entry to something like:

  `所选 KDF 使用不同的 PIM 参数，因此 VeraCrypt 不会重用当前的自定义 PIM。新的卷头信息将使用所选 KDF 的默认 PIM，除非您 在“新密码”区域中勾选“调整 PIM”并输入自定义值。\n\n您确定要继续吗？`

idrassi:
Thanks!

PR #1774, opened 2026-06-15
Update Corsican translation on 2026-06

Patriccollu:
Hello,

This is an update of **Corsican** (co) localization to take these commits into account:

 * https://github.com/veracrypt/VeraCrypt/commit/170dfa83eea5dc36c9dc3e7aa7786be39562bfe9 Linux/macOS: fix hidden volume FAT size limit
 * https://github.com/veracrypt/VeraCrypt/commit/d6220089ca1a7ff6259e3b0718c9bb502520905f Fix Unmount All access keys
 * https://github.com/veracrypt/VeraCrypt/commit/fcd430d65918ef825a4e7cd9fd4953e89820009a Increment version to 1.26.29. Update signed Windows drivers
 * https://github.com/veracrypt/VeraCrypt/commit/75857757fedf78b62d094f5268a948c22856245d Reset PIM defaults when changing volume KDF
 * https://github.com/veracrypt/VeraCrypt/commit/ceea6052d99ef1800fddcde5ac0ac6424177d37a Remove unused entry from XML language file following previous commit changes

Cheers,
Patriccollu.


idrassi:
Thanks but I have one remark: In `PIM_RESET_ON_KDF_CHANGE_CONFIRM`, the final part translates "custom value" as `valore predefinitu`, which seems to mean predefined value. Is there a better Corsican wording that is closer to "custom"?

idrassi:
Thanks for the quick update

Patriccollu:
Thank you for your review and for your remark which is entirely justified!

As the key contains 'custom PIM', 'default PIM' and 'custom value' I probably missed the point ;-)

PR #1776, opened 2026-06-15
macOS: parse hdiutil -plist output via CoreFoundation instead of string scanning

damianrickard:
## Problem

On macOS, mount/dismount device discovery (`CoreMacOSX::MountAuxVolumeImage` and `UpdateMountedVolumeInfo`) parses `hdiutil … -plist` output with a hand-rolled string scanner. It searches for `<key>dev-entry</key>` / `<key>mount-point</key>` / `<key>image-path</key>` and assumes the `<string>` value immediately follows the key and that entries appear in a fixed order — the code comment even notes it *"assumes the value follows the requested key."* That is fragile against any change in hdiutil's plist layout.

## Change

Parse the plist with the CoreFoundation property-list API (`CFPropertyListCreateWithData` + dictionary/array navigation). An RAII `CFHolder` guarantees `CFRelease` on every path.

Behavior is preserved:
- Prefer the `system-entities` entry carrying a `mount-point`, else fall back to the first `dev-entry`.
- Match disk images by normalized `image-path`.

CoreFoundation is already linked on macOS (via Cocoa), so there are no build-system changes.

## Testing

Built on Apple Silicon (arm64). Verified end-to-end: mounting parses `hdiutil attach -plist` correctly (virtual device resolved) and dismount parses `hdiutil info -plist` correctly (clean detach). Also exercised the no-match / empty-output paths.

### DIFF
--- a/src/Core/Unix/MacOSX/CoreMacOSX.cpp
+++ b/src/Core/Unix/MacOSX/CoreMacOSX.cpp
@@ -13,6 +13,8 @@
 #include <fstream>
 #include <stdio.h>
 #include <unistd.h>
+#include <vector>
+#include <CoreFoundation/CoreFoundation.h>
 #include <sys/param.h>
 #include <sys/ucred.h>
 #include <sys/mount.h>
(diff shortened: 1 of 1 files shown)


idrassi:
Thanks for the patch.  Using CoreFoundation property-list parser is a clear improvement over scanning the XML text manually.
Merging it.

PR #1791, opened 2026-06-21
macOS: honor the CheckFilesystem repair flag

### LINKED ISSUE #1789: macOS: "Repair Filesystem" is identical to "Check Filesystem" (repair flag ignored)
### Environment
- OS: macOS (all versions)
- VeraCrypt: current `master`

### Summary
On macOS, the **Check Filesystem** and **Repair Filesystem** menu items do
exactly the same thing: they both just launch Disk Utility.app. The
`repair` argument is silently ignored, so there is no way to trigger an
actual repair pass from the UI, and the two menu items are
indistinguishable in behavior.

### Details
Both menu items funnel through `MainFrame::CheckFilesystem(bool repair)`,
which calls `Core->CheckFilesystem(selectedVolume, repair)`:

- `OnCheckFilesystemMenuItemSelected`  → `CheckFilesystem()`     → `repair = false`
- `OnRepairFilesystemMenuItemSelected` → `CheckFilesystem(true)` → `repair = true`

On Linux/BSD, `CoreUnix::CheckFilesystem` honors the flag — it runs the
system `fsck` and passes `-n` only when `repair == false`
(`src/Core/Unix/CoreUnix.cpp`).

The macOS override, however, ignores both the volume and the flag entirely
(`src/Core/Unix/MacOSX/CoreMacOSX.cpp`):

```cpp
void CoreMacOSX::CheckFilesystem (shared_ptr <VolumeInfo> mountedVolume, bool repair) const
{
    list <string> args;
    struct stat sb;

    if (stat("/Applications/Utilities/Disk Utility.app", &sb) == 0)
        args.push_back ("/Applications/Utilities/Disk Utility.app");
    else
        args.push_back ("/System/Applications/Utilities/Disk Utility.app");

    Process::Execute ("/usr/bin/open", args);
}
```

`mountedVolume` and `repair` are unused — Disk Utility is opened with no
target either way.

### Steps to reproduce
1. Mount a volume on macOS.
2. Right-click it → **Repair Filesystem**, and separately → **Check Filesystem**.
3. Both simply open Disk Utility.app with no volume selected and no
   verify/repair action taken; the choice between them has no effect.

### Expected
The two actions should differ, and "Repair Filesystem" should actually
attempt a repair on the mounted volume's filesystem — consistent with the
Linux/BSD behavior.

### Proposed fix (seeking direction before submitting a PR)
I have a working patch that runs `diskutil verifyVolume` / `repairVolume`
on the VeraCrypt virtual device according to the flag. Because the Core
layer has no GUI, it surfaces the diskutil output by writing a small
helper script to a securely-created temp file (per-user temp dir via
`mkstemps`, `O_EXCL`/`0600`, `fchmod 0700`, removed afterward) and opening
it in Terminal, falling back to Disk Utility.app when no virtual device is
available. The device path is strictly validated (`/dev/[r]diskN[sM]`)
before use.

Before I open a PR, I'd like your input on the approach, since it
introduces a new mechanism on macOS:
- Is launching the check/repair in a Terminal window acceptable, or would
  you prefer the result be surfaced differently (e.g. through the GUI
  layer, or the existing privileged service)?
- Any objection to the temp-script-in-Terminal pattern as described?

Happy to adjust the implementation to match your preference. Thanks!

damianrickard:
Fixes #1789.

### Problem

On macOS, **Check Filesystem** and **Repair Filesystem** did not honor the selected operation. `CoreMacOSX::CheckFilesystem()` ignored the `repair` flag and only launched `Disk Utility.app` without targeting the selected VeraCrypt volume. As a result, the UI could not directly run a check or repair on the mounted volume.

### Fix

Run `diskutil` against the mounted VeraCrypt virtual device:

- **Check Filesystem** runs `diskutil verifyVolume`
- **Repair Filesystem** runs `diskutil repairVolume`

The result is shown in a Terminal window through a temporary `.command` helper script. If no suitable virtual device is available, VeraCrypt falls back to launching `Disk Utility.app`.

On macOS, `CoreServiceProxy::CheckFilesystem()` now calls the local macOS implementation directly instead of forwarding the request through `CoreService`. This keeps the helper script in the GUI user context even after an elevated device-hosted mount, avoiding root-owned temporary scripts and unreadable/unusable Terminal sessions.

### Safety

- The device path is strictly validated as `/dev/diskN`, `/dev/rdiskN`, or the corresponding slice form before it is embedded in the helper script.
- The helper script is created in the per-user temporary directory using `mkstemps()`, then permissioned with `fchmod()` on the open descriptor.
- The script is removed through an `EXIT` trap, and VeraCrypt also unlinks it if launching Terminal fails.
- Separate user-facing messages explain the Check and Repair flows.

### Validation

- Built successfully with FUSE-T using `src/Build/build_veracrypt_macosx.sh -f`.
- Manually tested Repair Filesystem on an elevated device-hosted mount.
- Verified that the temporary `VeraCrypt-fsck-*.command` helper is removed after completion.

### DIFF
--- a/Translations/Language.ar.xml
+++ b/Translations/Language.ar.xml
@@ -1498,6 +1498,8 @@
     <entry lang="ar" key="LINUX_REMOUNT_BECAUSEOF_SETTING">يرجى ملاحظة أن أي أحجام مثبتة حاليًا تحتاج إلى إعادة تثبيتها قبل أن تتمكن من استخدام هذا الإعداد.</entry>
     <entry lang="ar" key="LINUX_UNKNOWN_EXC_OCCURRED">حدث استثناء غير معروف.</entry>
     <entry lang="ar" key="LINUX_FIRST_AID">"سيتم تشغيل “أداة القرص” بعد الضغط على "موافق".\n\nيرجى تحديد حجمك في نافذة أداة الأقراص والضغط على "التحقق من القرص" أو زر "إصلاح القرص" على صفحة "الإسعافات الأولية".</entry>
+    <entry lang="en" key="MACOSX_CHECK_FILESYS">A Terminal window will open after you press 'OK' and check the file system on the selected VeraCrypt volume using 'diskutil'. The result will be shown in that window.\n\nIf the check cannot be started, Disk Utility will be launched instead.</entry>
+    <entry lang="en" key="MACOSX_REPAIR_FILESYS">A Terminal window will open after you press 'OK' and attempt to repair the file system on the selected VeraCrypt volume using 'diskutil'. The result will be shown in that window.\n\nIf the repair cannot be started, Disk Utility will be launched instead.</entry>
     <entry lang="ar" key="LINUX_MOUNT_ALL_DEV">تركيب جميع الأجهزة</entry>
     <entry lang="ar" key="LINUX_ERROR_LOADING_CONFIG">خطأ أثناء تحميل ملفات التكوين الموجودة في </entry>
     <entry lang="ar" key="LINUX_SELECT_FREE_SLOT">يرجى اختيار فتحة محرك مجانية من القائمة.</entry>
--- a/Translations/Language.be.xml
+++ b/Translations/Language.be.xml
@@ -1498,6 +1498,8 @@
     <entry lang="be" key="LINUX_REMOUNT_BECAUSEOF_SETTING">Калі ласка, звярніце ўвагу, што любыя змантаваныя ў цяперашні час тамы неабходна перамантаваць, перш чым яны змогуць выкарыстоўваць гэты параметр.</entry>
     <entry lang="be" key="LINUX_UNKNOWN_EXC_OCCURRED">Адбылося невядомае выключэнне.</entry>
     <entry lang="be" key="LINUX_FIRST_AID">«Дыскавая ўтыліта будзе запушчана пасля націску «ОК».\n\nВыберыце свой том у акне Disk Utility і націсніце кнопку «Праверыць дыск» або «Адрамантаваць дыск» на старонцы «Першая дапамога».</entry>
+    <entry lang="en" key="MACOSX_CHECK_FILESYS">A Terminal window will open after you press 'OK' and check the file system on the selected VeraCrypt volume using 'diskutil'. The result will be shown in that window.\n\nIf the check cannot be started, Disk Utility will be launched instead.</entry>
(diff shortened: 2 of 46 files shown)


idrassi:
Thanks for the PR. The direction looks good but I see a couple of issues that should be addressed before merging.

My main concern is the elevated service case. On macOS `CheckFilesystemRequest` doesn't request elevation, but once VeraCrypt has started the elevated CoreService, later requests can still be served by that elevated process. In that case this code creates the temporary `.command` file as root with mode 0700, then opens it in Terminal. Terminal normally runs in the GUI user session, so the script may not be readable or executable by the user, especially after a privileged device-hosted mount.

Please handle this case explicitly and test after an elevated mount.

Also, the existing macOS message shown before CheckFilesystem still says Disk Utility will be launched and asks the user to choose Verify/Repair manually. With this change the normal path opens Terminal and runs diskutil automatically, so that message needs to be updated.

Minor cleanup: if open fails after the script is created, the fallback path leaves the temporary script behind. A trap-based cleanup in the script would also avoid stale files if the window is closed early.

damianrickard:
Thanks for the detailed review — all three points addressed in the updated commit.

**Elevated service case.** Good catch. You're right that although `CheckFilesystemRequest` reports `RequiresElevation() == false` on macOS, once a device-hosted mount has started the elevated core service every subsequent request is forwarded to that root process, so the helper script would be created as root (0700) and the Terminal opened in the GUI session couldn't read or execute it. Rather than special-casing the elevated path, I moved the check into the unprivileged application process: `CoreServiceProxy::CheckFilesystem` now calls the core implementation directly on macOS instead of sending a service request. The check never needs root (diskutil acts on the virtual device and prompts for its own authorization when required), so the script is always owned by the GUI user regardless of whether the elevated service is running.

I tested this after a privileged device-hosted mount (mounted an encrypted device-hosted volume, which started the elevated service, then ran Check Filesystem): the Terminal window opens in the user session, runs `diskutil verifyVolume` on the virtual device, and shows the result — no permission problems.

**Pre-check message.** Replaced the old macOS message (which still mentioned Disk Utility and asked the user to pick Verify/Repair manually) with two new check- and repair-specific strings describing the automatic diskutil flow.

**Stale script on failure.** The script now installs `trap 'rm -f "$0"' EXIT` so it is removed even if the window is closed early, and it is also `unlink`ed if launching Terminal fails.

I also added the two new strings to all the translation files (English placeholders) so the XML key-completeness check passes.

idrassi:
Thanks, this addresses the concerns I raised. 

Before merging, can you please confirm that you also tested the Repair Filesystem path after an elevated device-hosted mount, not only Check Filesystem? repairVolume is the more permission-sensitive case.

Also, please update the PR description: it still says this is a single-file change, and it still describes the helper-script path as potentially running elevated, which is no longer accurate after the latest update.

idrassi:
I could test the change and I confirmed Repair Filesystem path with device-hosted volume.
I also updated PR description.

Merging this PR. Thank you.

PR #1797, opened 2026-06-22
Update Polish translation

Matthaiks:


idrassi:
Thanks. The AI review proposed the following:

In MACOSX_CHECK_FILESYS, could you change "Jeśli nie uda się przeprowadzić sprawdzania" to "Jeśli nie uda się uruchomić sprawdzaniaé? This better matches the actual fallback condition, where VeraCrypt opens Disk Utility only if starting the diskutil check cannot be launched.

After that, this is ready to merge.

Matthaiks:
Done. Thanks for the review.

PR #1799, opened 2026-06-26
Update Norwegian Bokmål translation

sandakersmann:
Update Norwegian Bokmål translation

### DIFF
--- a/Translations/Language.nb.xml
+++ b/Translations/Language.nb.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <VeraCrypt>
   <localization prog-version="1.26.29">
-    <language langid="nb" name="Norsk Bokmål" en-name="Norwegian (Bokmål)" version="14-06-2026" translators="Marius Kjærstad" />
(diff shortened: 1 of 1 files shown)


idrassi:
Thanks. Below are two small wording points reported by the AI review:
  - For the macOS filesystem strings, please use the official Norwegian app name `Diskverktøy` instead of lowercase `diskverktøy`.
  - For `IDD_BENCHMARK_DLG`, `ytelsestest for algoritmer` may be better than `ytelsestest for algoritme`.

Apart from these minor points, this looks ready.

sandakersmann:
"algoritmer" is the plural form. Not sure what form is best to use here. You have any idea?

Should we use singular or plural form?

idrassi:
Indeed. I think the review was wrong on this.  The English source string is `Algorithms Benchmark` so singular `ytelsestest for algoritme` reads like a benchmark for one algorithm only.

What about the other point?

sandakersmann:
Fixed :)

idrassi:
Thanks

PR #1802, opened 2026-06-28
Update Language.ro.xml

TigerxWood:
1. Made some more corrections.
2. Also on line 1543 there is an 'Options &gt;' &gt; ... I did not find anywhere 'Options &gt;' is this maybe about key="IDC_MOUNT_OPTIONS">Mount Opti&amp;ons...</entry>?


idrassi:
Thanks. Below is review feedback:

Please keep the VeraCrypt name in `IDM_WEBSITE`, reconsider the repeated change from `doar-citire` to `doar în mod citire` because it is inconsistent and less natural, and reword the PIM help text so `defaults` is translated clearly, e.g. “valorile implicite”.

Regarding the `Options > ...` string: it refers to the menu path `Options > Mount partition using system encryption`. The Romanian `Opțiuni > Montare partiție folosind criptarea sistemului` is the right idea: the odd extra `>` comes from the English source and can be fixed separately.

TigerxWood:
1. I changed IDM_WEBSITE to be consistent with the right click from tray icon where is the same and also in another two places, thinking that is obvious that web page is about Veracrypt. 
224    key="IDM_HOMEPAGE">&amp;Homepage </entry>
386    key="MENU_WEBSITE">   Home&amp;page   </entry>
1000  key="HOMEPAGE">Homepage</entry>

2. About doar în mod citire is a better choice.
3. PIM help text I will change it again
3. I still don't find that Options > anywhere in the entire file... maybe you could show me the row number

idrassi:
Thanks for the clarification.

So, I checked manually:

For `IDM_WEBSITE`: I understand the consistency argument with `IDM_HOMEPAGE`, `MENU_WEBSITE`, and `HOMEPAGE`, but the English source intentionally distinguishes them. Those keys are `Homepage`, while `IDM_WEBSITE` is `VeraCrypt Website`. Also, many other translations keep the product name for `IDM_WEBSITE`. Please use something like `Site-ul VeraCrypt` or `Site web VeraCrypt` there.

For `doar în mod citire`: if you prefer changing the wording, please make the ReadOnly terminology consistent across the Romanian file. Right now `READ_ONLY` and other strings still use `doar-citire`, while this PR changes only some sentences to `doar în mod citire`.

For the PIM help text: thanks.

For `Options >`: it is in the English source file, not literally in the Romanian translation. See `src/Common/Language.xml`, line 1538: 
`'Options &gt;' &gt; 'Mount partition using system encryption'`.

In the Romanian file it corresponds to line 1538:
`„Opțiuni” &gt; „Montare partiție folosind criptarea sistemului”`.

So there is nothing specific to fix in Romanian for that point; the odd extra `>` comes from the English source and can be handled separately.

TigerxWood:
Well, as I said from the start the only string where 'Options &gt;' &gt;  appears is in that string from line 1538. So where is the actual meniu option with "'Options &gt;'" string, because there is nowhere. The only button that goes to 'Mount partition using system encryption' oprtion that ai found is in the line 152 key="IDC_MOUNT_OPTIONS">Mount Opti&ons... so maybe the string from line 1538 should be changed to : 'Mount Opti&ons...' &gt; 'Mount partition using system encryption'.
I forgot to update IDM_WEBSITE but I will do after we resolve this issue with options

idrassi:
My bad: VeraCrypt appends ` >` when the options panel is collapsed at runtime!

idrassi:
The relevant control is not `IDC_MOUNT_OPTIONS`: that is the Windows “Mount Options...” button. For this Linux message, the user opens the `Options` panel in the password/mount dialog and then selects the checkbox whose label is `LINUX_MOUNT_SYSTEM_ENC_PREBOOT`.

On my side, I pushed a commit to fix the English source to avoid the malformed `Options >' >` wording.
A more explicit wording would be: “by clicking ‘Options’ and selecting ‘Mount partition using system encryption’”.

For Romanian, please translate that meaning rather than referencing `IDC_MOUNT_OPTIONS`.

TigerxWood:
ok, but in the key="LINUX_MOUNT_SYSTEM_ENC_PREBOOT">Mount partition &amp;using system encryption (preboot authentication) there is a '(preboot authentication)' additional string should I translate as it is, ...and selecting 'Mount partition using system encryption (preboot authentication)' ?

idrassi:
Yes, please translate the parenthetical because it's part of the Linux checkbox label.

In the Linux UI, the user clicks `Options` to expand the password dialog, then selects the checkbox `Mount partition using system encryption (pre-boot authentication)`.

I have pushed a commit to update the English source to make this explicit.

For Romanian, please use the same translated label in both places, for example:

`LINUX_MOUNT_SYSTEM_ENC_PREBOOT`:
`Montare partiție folosind criptarea sistemului (autentificare la pre-pornire)`

and in `LINUX_EX2MSG_PASSWORDORMOUNTOPTIONSINCORRECT`:
`... puteți face acest lucru apăsând pe „Opțiuni” și selectând opțiunea „Montare partiție folosind criptarea sistemului (autentificare la pre-pornire)”.`

idrassi:
Thank you for your help and patience!

PR #1806, opened 2026-06-29
fix: clarify outer-vs-hidden credentials in hidden volume protection UI

### LINKED ISSUE #1673: Hidden volume protection dialog is misleading about required password (outer vs hidden)
Summary:
When mounting a hidden volume with “Protect hidden volume against damage” enabled in the Windows GUI,
the password dialog is misleading and does not clearly state that it requires the OUTER (decoy) volume
credentials. This leads to common authentication failures (Error 9135) even for correctly created volumes.

Environment:
- VeraCrypt version: (1.26.24)
- OS: Windows (Windows 11)
- Volume type: File container
- Outer and hidden volumes use exFAT
- Outer and hidden volumes use completely random and distinct passphrases
- Outer volume: 20GB [AES+SHA512+default PIM(0), no key files]
- Inner hidden volume: 15GB [AES+SHA512+PIM(485)+keyfile]

Problem description:
The dialog shown when enabling “Protect hidden volume against damage” refers to “password” and mentions
that a cached password may be used, but it does not explicitly state that the OUTER volume password
(and PIM/keyfiles used at outer-volume creation) are required.

Because the dialog looks identical to a normal mount dialog, it is very natural for users to re-enter
the hidden volume password, hidden PIM, or hidden keyfiles. This results in Error 9135, which can
incorrectly suggest corruption or design errors—even immediately after volume creation.

Observed behavior:
- Hidden volume mounts successfully when protection is disabled
- Enabling protection causes Error 9135 unless outer volume credentials are entered
- Error message does not explain that the wrong volume’s credentials were provided

Expected behavior / suggestion:
- Dialog should explicitly state: “Enter OUTER (decoy) volume password. Do NOT enter hidden volume password here.”
- Consider labeling fields as “Outer volume password / PIM / keyfiles”
- Optionally clear PIM and keyfile selections when the dialog opens to prevent accidental reuse

Impact:
This wording ambiguity affects even technically experienced users and creates unnecessary confusion
around hidden volumes, which are already a sensitive and high‑stakes feature.

mvanhorn:
## Summary
Reword the English string entries so the credential split is explicit: clarify the Mount Options "Hidden Volume Protection" section (IDT_HIDDEN_VOL_PROTECTION / IDT_HIDDEN_PROT_PASSWD) to state it expects the HIDDEN volume's password, and add a short clarifying note that the main password dialog must receive the OUTER volume password when protection is enabled. In src/Mount/Mount.c, in the Mount Options dialog handling (MountOptionsDlgProc, around the IDC_PROTECT_HIDDEN_VOL / IDT_HIDDEN_PROT_PASSWD handling) and where the mount-failure error is surfaced, conditionally present the clarified wording when ProtectHiddenVolume is set so the failure message explains the two credential sets rather than the generic PASSWORD_OR_KEYFILE_WRONG text. Adjust the corresponding static-control wording/sizing in src/Mount/Mount.rc only if the longer label requires it (text is loaded from Language.xml via LocalizeDialog, so most changes are string-only).

## Why this matters
When mounting with "Protect hidden volume against damage" enabled in the Windows GUI, the main password dialog and the Mount Options protection section do not make clear which credential set goes where. The main dialog needs the OUTER (decoy) volume password while the Mount Options "Hidden Volume Protection" section needs the HIDDEN volume password, but the labels and the wrong-password error use generic text. This causes avoidable authentication failures (the reporter hit Error 9135) for correctly created volumes. The maintainer (idrassi) confirmed the requested behavior change (making the protection section ask for the outer password) would be incorrect, but explicitly agreed there is "a real UI ambiguity" and stated "We will consider improving the Windows GUI labels/error text to make this explicit." This plan implements only that maintainer-endorsed text/label clarification, not any behavior change.

See https://github.com/veracrypt/VeraCrypt/issues/1673.

## Testing
- Happy path: enable "Protect hidden volume against damage", observe the protection section clearly states it needs the HIDDEN volume password and the main dialog clearly indicates the OUTER volume password is required; mounting the outer volume with protection succeeds as before. - Edge case: protection disabled - labels and dialogs render with original wording, no clarifying note shown, no layout regression. - Edge case: verify translated builds still load (non-English language files fall back to English keys; no missing-key crash) and the new/edited keys exist for all referenced controls. - Error path: enter wrong outer/hidden credentials with protection enabled and confirm the failure message now distinguishes the two credential sets instead of the generic incorrect-password text.

Fixes #1673


idrassi:
Thank you for the PR. The direction is correct: hidden volume protection must continue to mount the outer volume while using the hidden volume credentials only to determine the protected range.

However, I can't merge this as-is because the new error message is currently only used in the “Auto-Mount All Devices” failure path, even though the PR description says the mount-failure error is clarified generally. The normal single volume mount path still reports ERR_PASSWORD_WRONG through `GetWrongPasswordErrorMessage`, so the main case described in #1673 would still show the generic password/keyfile/PIM/KDF error.

Also, the main dialog clarifies only the password label. Since the ambiguity also applies to PIM, KDF/PRF, and keyfiles, we should either add a clear note that the main dialog expects outer volume credentials, or conditionally clarify those labels too.

One style point: newly added entries in Language.xml should be appended at the bottom of the existing entries, not inserted in the middle. This makes it easier for translators to track new strings. Please move the newly added keys there.

Please update the PR accordingly.

idrassi:
@mvanhorn did you have a chance to look into my feedback?

mvanhorn:
Thanks for the detailed review - all three addressed:

1. The clarified message now also shows on the single-volume mount path, not just Auto-Mount: added a `GetWrongPasswordErrorMessageEx` variant that emits `HIDVOL_PROT_PASSWORD_OR_KEYFILE_WRONG` when hidden volume protection is active, wired into the `ERR_PASSWORD_WRONG` handling in `MountVolume`.

2. For the PIM/PRF/keyfile ambiguity, the main dialog now shows an info balloon on the password field noting it expects the outer (decoy) credentials when protection is enabled, rather than only relabeling the password label.

3. Moved the newly added Language.xml entries to the bottom of the entries block.

I don't have a Windows toolchain to build here, so a look on your end would be appreciated.

idrassi:
Thank you for the update. The direction is improved, but the following issues still need to be addressed before merge:

  1. XML validation is failing. The three new IDs are absent from all 42 translation files. English fallback prevents a crash, but produces mixed language dialogs. Please add the required entries to every translation file.

  2. GetWrongPasswordErrorMessageEx skips TCBootLoaderOnInactiveSysEncDrive when hidden volume protection is enabled. This removes the wrong mount mode message and SYSENC_MOUNT_WITHOUT_PBA_NOTE from an important hidden system path. Please combine that guidance with the new credential message.

  3. The balloon is not reliable as the main explanation. VeraCrypt hides edit balloons during typing and paste, focus may remain on the Mount Options button, and the mechanism is also used for Caps Lock. Please use persistent text or labels that clearly map the password, PIM, KDF, and keyfiles for both volumes.

  4. The direct MessageBoxW call bypasses the established Silent handling. Please use the common message path.

  5. Please restore IDT_HIDDEN_PROT_PASSWD and IDT_HIDDEN_VOL_PROTECTION to their original positions and append only the new entries. Also use "outer volume" instead of "decoy volume", use KDF consistently, and avoid "selected volume" in the automatic device mounting message.

This PR affects many use cases, so comprehensive testing is needed in addition to the code changes.
Please describe the scope of testing you performed, including the platforms and scenarios covered, so I can assess what remains before the PR can be accepted.

mvanhorn:
Items 1, 2, 4, and 5 are in 890c264:

- Restored IDT_HIDDEN_PROT_PASSWD and IDT_HIDDEN_VOL_PROTECTION to their original position and text so their existing translations stay valid; the clarification now lives only in the new keys. Switched to "outer volume" throughout (dropped "decoy"), used KDF consistently, and reworded the wrong-credentials string to avoid "selected volume".
- GetWrongPasswordErrorMessageEx now appends SYSENC_MOUNT_WITHOUT_PBA_NOTE in the hidden-volume-protection path as well, keeping the wrong-mode message for the non-hidden path only, so the hidden system path retains that guidance alongside the credential message.
- Replaced the direct MessageBoxW with WarningDirect so it goes through the common Silent-aware path.
- Added the three new keys to all 42 translation files with the English text so XML validation passes; localization can follow.

I have not done item 3 yet (replacing the edit balloon with persistent labels), because I want to get it right. My plan is persistent static labels near each field mapping password/PIM/KDF/keyfiles for the outer vs hidden volume instead of the balloon. Does that match what you had in mind?

On testing scope, to be upfront: I could not build or run this on Windows in my environment, so I have not exercised the dialog end to end. What I verified is XML well-formedness of all 43 language files and the C changes by review. Before merge I would want the hidden-volume-protection mount paths tested on Windows (outer-vs-hidden credential handling, the SysEnc inactive-drive path, and the message routing), and I am glad to script specific scenarios. Happy to hold the balloon work and that testing until we settle item 3.

mvanhorn:
This is a wording-only change — it reworks the English strings (via Language.xml / LocalizeDialog) and conditionally surfaces the clarified text in the Mount Options dialog and the mount-failure message when Protect Hidden Volume is enabled; it doesn't touch mount or crypto logic. The surface to confirm is the Windows GUI flow: mounting with "Protect hidden volume against damage" enabled and the wrong-credentials failure path. If you tell me which platforms/scenarios you want covered, I'll run through them and report back.

mvanhorn:
The listed issues are handled in 3c0f42a across all 43 language files, checked against the full key set with escape validation. Let me know if anything else stands out.

PR #1808, opened 2026-07-01
Windows: bound lower device I/O waits

### LINKED ISSUE #1807: Windows 11 kernel storage hang: Bitdefender bdservicehost.exe stalls on pending write to VeraCryptVolumeY during exFAT USB-volume transfer
I experienced two related storage-instability patterns involving VeraCrypt and Bitdefender.

First, after enabling VeraCrypt system encryption on the Windows system drive, the computer repeatedly crashed during boot or restart with SYSTEM_THREAD_EXCEPTION_NOT_HANDLED (0x7E) BSODs naming veracrypt.sys. A separate PAGE_FAULT_IN_NONPAGED_AREA (0x50) crash occurred in the Windows Filter Manager / NTFS I/O path while Bitdefender and VeraCrypt were both active.

Later, after permanently decrypting the Windows system drive, I reproduced a serious kernel storage hang while Bitdefender was active and a sustained file transfer was running between two external VeraCrypt volumes. The affected VeraCrypt-mounted exFAT volume became inaccessible, VeraCrypt stopped responding, Bitdefender Virus Shield terminated, and Windows could not restart normally.

The live-kernel dump showed that a bdservicehost.exe thread owned the blocked exFAT resource while waiting for a pending write whose current driver was \Driver\veracrypt and whose target device was VeraCryptVolumeY.

I am reporting this as a possible compatibility or interaction issue involving VeraCrypt, Bitdefender filesystem filtering, exFAT, and sustained USB storage I/O. I am not asserting that VeraCrypt alone or Bitdefender alone is conclusively the sole cause.

# Expected behavior
Two mounted VeraCrypt volumes on external USB drives should remain accessible and a sustained file transfer between them should complete normally.

VeraCrypt should remain responsive, Windows Explorer should remain responsive, and Windows should be able to shut down or restart normally.

# Observed behavior
During a sustained file transfer between two external USB drives containing VeraCrypt volumes, the mounted VeraCrypt volume Y: became inaccessible while the physical USB drive still appeared present in Windows.

At the same time:

VeraCrypt became unresponsive.
Windows Explorer became affected by the inaccessible volume.
Bitdefender Antivirus / Virus Shield stopped responding and Windows Security reported that antivirus protection needed attention.
Bitdefender’s Virus Shield service later logged as having terminated unexpectedly.
Bitdefender file-system filters unloaded and reloaded.
Windows generated a ResourceTimeout live-kernel dump.
Normal Windows restart remained stuck on the spinning-dots screen and eventually required a forced shutdown.

There was no normal BSOD during this particular incident.

WinDbg analysis of the ResourceTimeout live dump showed:

EXRESOURCE_TIMEOUT_LIVEDUMP (0x1CC)
The timed-out exFAT resource was owned by a thread in bdservicehost.exe.
That Bitdefender-owned thread was waiting in an exFAT write/flush path.
Its active write IRP was IRP_MJ_WRITE.
The current driver for that write IRP was \Driver\veracrypt.
The target device object was identified as:
VeraCryptVolumeY \Driver\veracrypt
An explorer.exe thread was blocked waiting to acquire the same exFAT resource.

Relevant WinDbg excerpt:

IRP_MJ_WRITE
\Driver\veracrypt
exfat!FppSingleSyncCompletionRoutine

Device object:
VeraCryptVolumeY \Driver\veracrypt

I am not asserting that VeraCrypt alone caused this. The available live dump does not show enough of the lower USB/storage stack to determine whether the initiating problem was VeraCrypt, Bitdefender, their interaction, exFAT, USB storage, the enclosure/cable, or another lower storage component.

However, VeraCryptVolumeY was directly identified in the pending write path during the kernel resource timeout.

Additional context: earlier on this same laptop, after VeraCrypt system encryption was enabled on the Windows system drive, I experienced repeated SYSTEM_THREAD_EXCEPTION_NOT_HANDLED (0x7E) BSODs naming veracrypt.sys during boot/restart. I also experienced a separate PAGE_FAULT_IN_NONPAGED_AREA (0x50) event involving Filter Manager while Bitdefender was active. I have since permanently decrypted the Windows system drive.

# Earlier VeraCrypt system-encryption BSOD history
Before the external-volume ResourceTimeout incident, I had enabled VeraCrypt system encryption on the Windows system drive on this same laptop.

After system encryption was completed, I experienced repeated boot and restart BSODs, including multiple:

SYSTEM_THREAD_EXCEPTION_NOT_HANDLED (0x7E)

events that identified:

veracrypt.sys

as the named driver.

These failures occurred during normal restart or boot activity, including cases where Windows could not start normally after the VeraCrypt pre-boot authentication stage. At least one crash did not produce a usable dump, but the repeated BSOD screen reports consistently named veracrypt.sys.

A separate kernel-memory dump from this period recorded:

PAGE_FAULT_IN_NONPAGED_AREA (0x50)

The active process was bdservicehost.exe, and the stack was in the Windows Filter Manager / NTFS I/O path, including:

FLTMGR!FltpPerformPreCallbacksWorker

At the time of that crash, VeraCrypt was loaded and Bitdefender's filesystem-filter stack was active, including vlflt, BdDci4, Ignisv2, Gemma, atc, and Trufos.

The crash stack did not directly show a Bitdefender driver function or veracrypt.sys executing at the faulting instruction. Therefore, I cannot conclude from that crash alone that either product was the sole cause. However, the event occurred in a heavily filtered filesystem I/O environment with both products active.

After I permanently decrypted the Windows system drive, I completed several normal restarts without a BSOD. This suggested that VeraCrypt system encryption was an important part of the earlier failure pattern, but it did not prove that normal VeraCrypt mounted volumes were stable.

The later ResourceTimeout incident then occurred with the Windows system drive decrypted, while Bitdefender was active and ordinary external VeraCrypt volumes were being used under sustained I/O. This indicates that the instability is not limited only to VeraCrypt system encryption and may involve an interaction between Bitdefender, VeraCrypt, the filesystem-filter stack, and external USB/exFAT storage.

# Steps to reproduce
Start Windows normally with Bitdefender Antivirus installed and active.
Connect two external USB drives containing VeraCrypt-encrypted volumes.
Mount both VeraCrypt volumes. In the incident, the affected mounted volume was Y: and used exFAT.
Start a sustained large file transfer between the two mounted VeraCrypt volumes.
Leave the transfer running.

Observed result in my case: after the transfer had been running, the Y: mounted VeraCrypt volume became inaccessible, VeraCrypt stopped responding, Bitdefender Virus Shield failed, Windows created a ResourceTimeout live dump, and Windows could not restart normally.

I understand this may not reproduce on every attempt or every computer. The issue occurred on the hardware and software environment listed below.

# Screenshots

# Your Environment

VeraCrypt version: 1.26.29

Operating system and version: Windows 11 Home, version 25H2, OS build 26200.837

System type: 64-bit operating system, x64-based processor

Laptop: ASUS ROG Strix G16 G615LW

BIOS version: G615LW.336

CPU: Intel Core Ultra 9 290HX Plus

Memory: 64 GB DDR5, 5600 MT/s

Bitdefender: 27.0.60.339

GPU: NVIDIA GeForce RTX 5080 Laptop GPU
Drive configuration: Windows system drive was decrypted at the time of this incident. VeraCrypt system encryption was not active. Two external USB drives containing VeraCrypt volumes were mounted.

Filesystem involved: exFAT on the affected VeraCrypt-mounted volume.

michaelsam94:
## Summary
- replace the unbounded synchronous lower-device read/write wait with an owned asynchronous IRP path
- add a 120-second timeout that logs the stalled lower-device read/write, cancels the IRP, and returns STATUS_IO_TIMEOUT
- keep the original caller buffer safe after timeout by using a nonpaged bounce buffer and retaining cleanup context until the lower IRP completes

## Rationale
Issue #1807 reports a Windows storage hang where a thread is blocked on a pending write whose current driver is \Driver\veracrypt. The previous helper waited indefinitely for lower-device reads/writes, so a stalled USB/storage/filter-stack request could keep VeraCrypt's queue worker blocked indefinitely as well.

This does not assert that VeraCrypt alone causes the reported Bitdefender/exFAT/USB interaction. It makes VeraCrypt fail the lower I/O upward after a bounded wait instead of waiting forever.

Fixes #1807

## Testing
- git diff --check
- Not run: Windows driver build/runtime testing is not available in this macOS workspace.

### DIFF
--- a/src/Driver/Ntdriver.c
+++ b/src/Driver/Ntdriver.c
@@ -3864,31 +3864,114 @@ void TCCloseFsVolume (HANDLE volumeHandle, PFILE_OBJECT fileObject)
 }
 
 
+#define TC_LOWER_DEVICE_IO_TIMEOUT_SECONDS 120
+
+typedef struct
+{
+	PDEVICE_OBJECT DeviceObject;
+	PIRP Irp;
+	PVOID Buffer;
+	KEVENT CompletionEvent;
+	IO_STATUS_BLOCK IoStatus;
+	volatile LONG Completed;
+	volatile LONG RefCount;
+} TC_LOWER_DEVICE_IO_CONTEXT;
+
+
+static VOID TCReleaseLowerDeviceIoContext (TC_LOWER_DEVICE_IO_CONTEXT *context)
+{
+	if (InterlockedDecrement (&context->RefCount) == 0)
+	{
+		IoFreeIrp (context->Irp);
+		ObDereferenceObject (context->DeviceObject);
+		TCfree (context->Buffer);
+		TCfree (context);
+	}
+}
+
+
+static NTSTATUS TCLowerDeviceIoCompletion (PDEVICE_OBJECT deviceObject, PIRP irp, PVOID completionContext)
+{
+	TC_LOWER_DEVICE_IO_CONTEXT *context = (TC_LOWER_DEVICE_IO_CONTEXT *) completionContext;
+
+	UNREFERENCED_PARAMETER (deviceObject);
+
+	context->IoStatus = irp->IoStatus;
+	InterlockedExchange (&context->Completed, TRUE);
+	KeSetEvent (&context->CompletionEvent, IO_DISK_INCREMENT, FALSE);
+	TCReleaseLowerDeviceIoContext (context);
+
+	return STATUS_MORE_PROCESSING_REQUIRED;
+}
+
+
 static NTSTATUS TCReadWriteDevice (BOOL write, PDEVICE_OBJECT deviceObject, PVOID buffer, LARGE_INTEGER offset, ULONG length)
 {
 	NTSTATUS status;
-	IO_STATUS_BLOCK ioStatusBlock;
+	TC_LOWER_DEVICE_IO_CONTEXT *context;
 	PIRP irp;
-	KEVENT completionEvent;
+	LARGE_INTEGER waitTimeout;
+	ULONG_PTR bytesToCopy;
 
 	ASSERT (KeGetCurrentIrql() <= APC_LEVEL);
 
-	KeInitializeEvent (&completionEvent, NotificationEvent, FALSE);
-	irp = IoBuildSynchronousFsdRequest (write ? IRP_MJ_WRITE : IRP_MJ_READ, deviceObject, buffer, length, &offset, &completionEvent, &ioStatusBlock);
-	if (!irp)
+	if (length == 0)
+		return STATUS_SUCCESS;
+
+	context = TCalloc (sizeof (TC_LOWER_DEVICE_IO_CONTEXT));
+	if (!context)
 		return STATUS_INSUFFICIENT_RESOURCES;
 
-	ObReferenceObject (deviceObject);
-	status = IoCallDriver (deviceObject, irp);
-
-	if (status == STATUS_PENDING)
+	context->Buffer = TCalloc (length);
+	if (!context->Buffer)
 	{
-		status = KeWaitForSingleObject (&completionEvent, Executive, KernelMode, FALSE, NULL);
-		if (NT_SUCCESS (status))
-			status = ioStatusBlock.Status;
+		TCfree (context);
+		return STATUS_INSUFFICIENT_RESOURCES;
 	}
 
-	ObDereferenceObject (deviceObject);
+	if (write)
+		memcpy (context->Buffer, buffer, length);
+
+	KeInitializeEvent (&context->CompletionEvent, NotificationEvent, FALSE);
+	context->RefCount = 2;
+	context->DeviceObject = deviceObject;
+	ObReferenceObject (deviceObject);
+
+	irp = IoBuildAsynchronousFsdRequest (write ? IRP_MJ_WRITE : IRP_MJ_READ, deviceObject, context->Buffer, length, &offset, &context->IoStatus);
+	if (!irp)
+	{
+		ObDereferenceObject (deviceObject);
+		TCfree (context->Buffer);
+		TCfree (context);
+		return STATUS_INSUFFICIENT_RESOURCES;
+	}
+
+	context->Irp = irp;
+	IoSetCompletionRoutine (irp, TCLowerDeviceIoCompletion, context, TRUE, TRUE, TRUE);
+
+	status = IoCallDriver (deviceObject, irp);
+
+	if (!InterlockedCompareExchange (&context->Completed, FALSE, FALSE))
+	{
+		waitTimeout.QuadPart = -((LONGLONG) TC_LOWER_DEVICE_IO_TIMEOUT_SECONDS * 1000 * 10000);
+		status = KeWaitForSingleObject (&context->CompletionEvent, Executive, KernelMode, FALSE, &waitTimeout);
+		if (status == STATUS_TIMEOUT)
+		{
+			Dump ("Lower device %s timeout: offset=%I64d length=%u\n", write ? "write" : "read", offset.QuadPart, length);
+			IoCancelIrp (irp);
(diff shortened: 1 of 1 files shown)


idrassi:
Thank you for digging into #1807 and putting together a concrete proposal. The idea of avoiding an indefinitely blocked VeraCrypt worker thread when a lower storage stack is wedged is valid, and the bounce-buffer/refcount direction shows real care.

After reviewing the PR and the regression reports, I don't think this is the right fix to merge for #1807.

My current assessment is that the 1.26.29 regression is more likely tied to recent driver I/O dispatch/queueing changes that landed after 1.26.24, especially:

- `a7ebddc5`: the change that moved IRP completion dispatch to `CriticalWorkQueue` and deferred some error/cancel completion paths;
- `d1f73ce4`: the ordered flush barrier change that routes `IRP_MJ_FLUSH_BUFFERS` through `EncryptedIoQueue`.

I have pushed a more conservative fix to `master`: revert the completion-dispatch change back toward the previous stable model (https://github.com/veracrypt/VeraCrypt/commit/c947e56b6efcf00ce7f03a919cb335457c2eb1c8), and keep ordered flush barriers behind an opt-in driver configuration flag, defaulting them off (https://github.com/veracrypt/VeraCrypt/commit/f6544a5b07c85bbd9e8af63e960207e99cd81c68). That reduces the regression surface on the affected paths while we collect more feedback.

The main reason I don't think this PR covers #1807 is that the dump points to a pending `IRP_MJ_WRITE` on a mounted non-system volume. For mounted non-system volumes, the backing I/O goes through the host file handle path in `EncryptedIoQueue.c`, for example:

```c
request->Item->Status = ZwWriteFile (
    queue->HostFileHandle,
    NULL,
    NULL,
    NULL,
    &ioStatus,
    request->Data,
    request->Length,
    &request->Offset,
    NULL);
```

`TCReadWriteDevice`, which this PR rewrites, is used on the filter/raw-device helper paths, not on that normal mounted-volume ZwWriteFile path. So this adds a substantial new mechanism without directly addressing the path that appears to hang in #1807.

Even as a separate hardening change, there are several issues that would need to be resolved first:

1. The context is not fully initialized. In the Windows driver, TCalloc() uses ExAllocatePoolUninitialized(), so fields such as Completed and IoStatus contain undefined data unless explicitly initialized. The PR reads these fields, which can skip the wait, return an undefined status, or copy read data before the lower IRP has completed.

2. The timeout/completion handoff needs a real state machine. After KeWaitForSingleObject() returns STATUS_TIMEOUT, the lower IRP can still complete before or during IoCancelIrp(). In that race, the caller can return STATUS_IO_TIMEOUT even  though the I/O completed. For writes, this ambiguity is especially risky.

3. The driver-created asynchronous IRP cleanup is incomplete. For direct-I/O devices, `IoBuildAsynchronousFsdRequest` can create and lock an MDL. Since the completion routine returns STATUS_MORE_PROCESSING_REQUIRED and the PR frees the IRP itself, the MDL must be unlocked and freed explicitly before IoFreeIrp().

4. IoCancelIrp() does not guarantee that a wedged storage IRP will complete. If the lower stack never completes the IRP,  the context, bounce buffer, IRP, and any associated MDL must remain allocated forever to avoid use-after-free. So this can unblock the VeraCrypt thread, but it does not actually recover the lower I/O.

5. The bounce-buffer design adds allocations and full-length copies to every affected lower-device I/O. On the filter path, that is the hot disk I/O path, so it would be a performance and NonPagedPool pressure regression.

I'm going to close this PR rather than merge it for #1807. A bounded lower-device wait mechanism may still be worth exploring later as a separate hardening change, but it would need a design that handles initialization, MDL cleanup, cancellation/completion ownership, write semantics, and performance impact very carefully.

Thanks again for the investigation and for taking the time to propose a concrete fix.

PR #1809, opened 2026-07-01
Remove triple punctuation

sandakersmann:
A lot of the triple punctuation do not add to the strings, and they are not uniform throughout. If this looks good to you I will start working on all the translations as well.

### DIFF
--- a/src/Common/Language.xml
+++ b/src/Common/Language.xml
@@ -8,9 +8,9 @@
     <font lang="en" class="title" size="21" face="Times New Roman" />
     <entry lang="en" key="IDCANCEL">Cancel</entry>
     <entry lang="en" key="IDC_ALL_USERS">Install &amp;for all users</entry>
-    <entry lang="en" key="IDC_BROWSE">Bro&amp;wse...</entry>
+    <entry lang="en" key="IDC_BROWSE">Bro&amp;wse</entry>
     <entry lang="en" key="IDC_DESKTOP_ICON">Add VeraCrypt icon to &amp;desktop</entry>
-    <entry lang="en" key="IDC_DONATE">Donate now...</entry>
+    <entry lang="en" key="IDC_DONATE">Donate now</entry>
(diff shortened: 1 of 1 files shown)


sandakersmann:
@idrassi What do you think about this concept?

amail80:
The trailing ellipsis on these labels is not punctuation but a user interface convention: it signals that activating the control opens a dialog requiring further input, rather than performing the action immediately. It is specified in the Windows, macOS and GNOME human interface guidelines, and it is how users tell `Save` from `Save As…` without reading any documentation.
 
Every affected key here does open a dialog:
 
| key | opens |
| --- | --- |
| `IDC_BROWSE`, `IDC_BROWSE_FILES`, `IDC_BROWSE_DIRS` | file/folder picker |
| `IDC_SELECT_FILE`, `IDC_SELECT_DEVICE` | volume selection dialog |
| `IDC_KEY_FILES`, `IDC_KEYFILES`, `IDC_NEW_KEYFILES` | keyfile manager |
| `IDC_SELECT_PKCS11_MODULE` | library picker |
| `IDC_MOUNT_OPTIONS`, `IDC_MORE_SETTINGS` | options dialog |
 
After this change `Mount Options` reads exactly like `Mount`, although one opens a dialog and the other acts at once. The same applies to `Select File` next to `Mount`, and to `Keyfiles` next to `Dismount`. The distinction is currently free — it costs three characters and no translation work — and removing it takes information out of the interface rather than tidying it up.
 
There is also a consistency issue: only the English strings are changed. The per-language `Language.*.xml` files keep their ellipses, so the same button would be labelled `Browse` in English and `Обзор…` in Russian, `Parcourir…` in French and so on. Bringing those in line would mean touching every translation.
 
If the intent was to remove genuinely stray punctuation — three dots typed as three periods rather than as a single ellipsis, or trailing dots on labels that do not open anything — that would be worth doing, but it is a different and much smaller set of strings than the one changed here.

PR #1823, opened 2026-07-10
Linux: fix tray icon menu build with INDICATOR=1

ElCruncharino:
## Problem

Building with `INDICATOR=1` fails on wxWidgets 3.2:

```
Forms/MainFrame.cpp:1630:134: error: cannot convert 'wxString' to 'const gchar*'
```

Four of the five tray menu labels in `ShowTaskBarIcon()` pass a `wxString` straight to `gtk_menu_item_new_with_label()`, which wants a `const gchar*`. The fifth (`SHOW_TC`/`HIDE_TC`) already calls `.mb_str()` and compiles fine. `INDICATOR` is opt-in and off by default, so perhaps this has gone unnoticed?

## Change

Add the missing `.mb_str()` to the other four labels. No behavior change, just makes it compile.

## Testing

Built with `make INDICATOR=1 WITHFUSE3=1` against wxWidgets 3.2.8 and libayatana-appindicator3. Confirmed the resulting binary links against libayatana-appindicator3 and shows a working tray icon.


idrassi:
Thanks for the PR.

I had not hit this locally with make INDICATOR=1 on several distributions, including with wxWidgets 3.2.8. That is expected on builds where wx still allows the unsafe implicit wxString => const char* conversion (wxUSE_UNSAFE_WXSTRING_CONV without wxNO_UNSAFE_WXSTRING_CONV). On configurations that disable that conversion, the four bare LangString[...] arguments fail exactly as you reported, while the existing `.mb_str()` on the show/hide item still compiles.

So this is a real portability fix (the inconsistency dates back to the original INDICATOR support in #815), not a general "wx 3.2 is broken" issue.

Could you confirm the distribution and whether your environment defines wxNO_UNSAFE_WXSTRING_CONV (or similar)? For the record, sibling PR #1822 mentions openSUSE Tumbleweed, is that the same setup?

Aside from that, the change looks good and matches the existing style. I will merge it.

ElCruncharino:
Confirmed on the exact box: openSUSE Tumbleweed, wxWidgets 3.2.8 (`gtk3-unicode-3.2`, official distro package).

I don't think its `wxNO_UNSAFE_WXSTRING_CONV`. That macro isn't defined anywhere in this package's `setup.h`, and I get the identical compile error whether I force it or not. `wxUSE_STL`, which this package builds with (`wxUSE_STL=1` → `wxUSE_STD_STRING_CONV_IN_WXSTRING=1`). In `wx/string.h`, the whole `operator const char*()`/`operator const void*()` block is wrapped in `#if !wxUSE_STD_STRING_CONV_IN_WXSTRING`, so in STL-mode builds that operator doesn't exist at all. `wxNO_UNSAFE_WXSTRING_CONV` doesn't appear to even come into play.

I rebuilt master with `make INDICATOR=1 WITHFUSE3=1` on this box and reproduced the exact 4 errors at the exact reported lines; applying this PR's diff fixed it and produced a working binary linked against `libayatana-appindicator3`.

